PR #6, opened 2022-08-04
fix for missing files

hsubramanianaks:


Tatsinnit (approved):

Tatsinnit (approved):
🎉❤️☕️🙏

src/vscode/src/.gitignore
```diff
@@ -0,0 +1,8 @@
+out
+dist
+node_modules
+.vscode-test/
+*.vsix
+
+# Parent .gitignore excludes 'debug' directories, but this one is legit
+!src/debug
```
Tatsinnit, line 8: 💡 just a not that we can remove this line, but we can do this latter.

hsubramanianaks: yes, we can remove once we move vscode to separate repo :)

elenavillamil (approved):

PR #17, opened 2022-08-15
delay cancelation token by one second

elenavillamil:
In some occasions we see token hang up error when disconnecting bridge debugger. This is caused because on disconnect cancellation signal to tear down server is sent immediately, sometime before response is sent to the client. Adding a delay on cancellation token so http server has enough time to send ok before being teared down.


hsubramanianaks (approved):
LGTM since this is keeping sync with VSTS repo and this will be last commit, going forward fixes will be done in OSS repo :) this is FYI everyone.

PR #19, opened 2022-08-16
Feature to release binaries

hsubramanianaks:
**Feature for releasing binaries via GitHub actions.**

- this feature releases binaries for three operating systems mac-OS, linux and windows.
- it also includes self-contained release ex includes the binaries required for bridge along with dotnet libraries, kubectl library.
- it also releases binaries as GitHub release with semantic versioning ex: 0.0.1
- Release to production happens only on creation of tag which is manual now so that we can keep control of release to production.
- Sample release from forked repo - https://github.com/hsubramanianaks/Bridge-To-Kubernetes/releases/tag/0.0.21


.github/workflows/release.yml
```diff
@@ -0,0 +1,131 @@
...
+      if: ${{ matrix.mode == 'win-x64' }}
+      with:
+        inlineScript: |
+          Copy-Item -Path ${{ github.workspace }}\src\EndpointManagerLauncher\bin\Release\netcoreapp3.1\win-x64\publish\ -Destination ${{ github.workspace }}\src\dsc\bin\Release\netcoreapp3.1\win-x64\publish\EndpointManagerLauncher -Recurse -Exclude **/*.pdb   
+        azPSVersion: '3.1.0'
+    - name: Download Kubectl
+      uses: Azure/powershell@v1
+      with:
+        inlineScript: |
+          if ('${{ matrix.mode }}' -eq 'win-x64') {
```
Tatsinnit, line 38: 💡 Cool, you could check this `setup-kubectl` - `Azure` GH action out :) https://github.com/Azure/setup-kubectl  More kind of Action based.

.gitignore
```diff
@@ -348,3 +348,6 @@ MigrationBackup/
 
 # Ionide (cross platform F# VS Code tools) working folder
 .ionide/
+
+#vscode launch and tasks json
+.vscode/
```
Tatsinnit, line 353: 💡 Shall we remove the `.vscode` dir structure from this repo? could be in separate PR since we are gitignoring it.

CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Changelog
+
+## [0.0.1] - 2022-08-16
```
Tatsinnit, line 3: 💡 Just thinking outloud: Is this the changelog versioning you guys are going with? Just thought is we want to keep it more basic, like: https://github.com/Azure/aks-periscope/blob/master/CHANGELOG.md#0010

Tatsinnit (commented):
💡 Thank you for adding me to this review, really appreciate it, added some thoughts. Please let me know how you feel. Thanks heaps.

.github/workflows/release.yml
hsubramanianaks: @Tatsinnit thank you for the suggestion, I know this GH action exists and checked it as well, but this downloads the kubectl library to the runner, two things we need the kubectl to be downloaded to specific path and also for windows it needs to download from different URL/blob, and it will be .exe file. potential improvement for this GH action which we can do as an enhancement and change it later. do you agree? I will check with Oliver as well on this.

.gitignore
hsubramanianaks: .vscode directory doesn't exist just added them to the .gitignore and directory is removed. are you saying we want this in separate PR ?

CHANGELOG.md
hsubramanianaks: I was just following this - https://keepachangelog.com/en/1.0.0/, with the date actually it is nicer to see when the change happened, also with #Added, # Changed, #Removed is better. If you are saying this is standard, we follow for all our Azure repo's we can adhere to this https://github.com/Azure/aks-periscope/blob/master/CHANGELOG.md#0010 , no worries at all.

.github/workflows/release.yml
hsubramanianaks: @Tatsinnit ignore my comment above, just found out this GH action takes care of downloading kubectl based on the OS I didn't know at the first place, I will update the code. Thanks. :) do you think I can do it as separate PR?

hsubramanianaks: But this job runs on windows-latest and downloads kubectl bundled with other dotnet libraries, bridge libraries for all three OS (linux, windows, OSX) in parallel. setup-kubectl GHA wouldn't do that, it auto detects the OS and downloads only for them. @OliverMKing this would be an enhancement which can be done later.

CHANGELOG.md
elenavillamil: I would rather to keep something with date on it. It makes live easier when debugging I think.  You can quickly see how old are things people are running with. As Hari mentioned if it is some Azure standard okay to change, but if not, I rather keep something with date :)

.gitignore
Tatsinnit: 💡☕️ Cool I meant this .vscode dir which currently exist in codebase: https://github.com/Azure/Bridge-To-Kubernetes/tree/main/src/.vscode , son of we are gitignoring then just bringing this dot in your attention. 😊

CHANGELOG.md
Tatsinnit: ☕️💡 cool, I guess my food for thought could be if we want to keep the current versioning strategy we should make it more visible/true to what we release  from change log version like for example : x.x.x.xxxxxx - dd.mm.yyhhmmss or whatever the versioning strategy this release will follow.

.gitignore
hsubramanianaks: Thank you @Tatsinnit missed checking in the main repo, I will remove them.

CHANGELOG.md
hsubramanianaks: So pipeline works here is screenshot 
image

Tatsinnit (approved):
❤️🙏☕️🥷

PR #24, opened 2022-08-21
Implement ability to execute command in Bridge environment shell

ajaffie:
Currently, any arguments passed into the connect command following '--' are stored in the `CommandExecuteArguments` property of the `CommandLineArgumentsManager`. Until now, this property was ignored.
Now `CommandExecuteArguments` are passed into `LaunchTerminalWithEnv()` whenever `performLaunch` is true, which then executes the provided command in the launched environment.

To avoid changing the way the shell is launched in each platform, the Windows method appends the command to the output script and the bash method executes the specified command instead of bash.

One thing I'm missing is how to change the help/usage output. I think it would be helpful if the usage looked something like this:
> Usage:  connect [options] [-- \<command\> [\<additional arguments\>...]]
> ...
> Command:
> &nbsp;&nbsp;Command to execute in the Bridge environment.
> Additional Arguments:
> &nbsp;&nbsp;Arguments to pass to the command launched in the Bridge environment.

I tried adding it as an argument but that didn't do what I wanted and setting the `Syntax` property in `ConnectCommand.Configure()` didn't have any effect at all:
```csharp
this._command.Syntax = $"{Environment.GetCommandLineArgs()[0]} {Name} [options] [-- [<command> [<command arguments>...]]";
```

I'd appreciate some help with this, I think this would be a really helpful feature.


elenavillamil:
For the help message updated, have you try updating description passed into "public void ConfigureHelp(string description, string extendedHelpText = "")"? You could update the description or provide extendedHelpText? Altough this may not be enough, the code is using [CommandLineApplication](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.commandlineutils.commandlineapplication?view=dotnet-plat-ext-1.1), which has a virtual [GetHelpText(String)](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.commandlineutils.commandlineapplication.gethelptext?view=dotnet-plat-ext-1.1#microsoft-extensions-commandlineutils-commandlineapplication-gethelptext(system-string)) method which if we add it as part of our command class we should be able to fully customize the help message. If you have more questions let me know, I can try applying your code later today or tomorrow and try as well.

ajaffie:
The extended help text parameter let me add text to the end and there was a switch that slightly modified the usage. Now the help output looks like this:

>  Redirect traffic from a service, deployment or pod running in your cluster to your local machine.
> 
> Usage:  connect [options] [[--] \<arg\>...]
> 
> Options:
>   -h|--help                                       Show help information
>   --container                                     The name of the container to redirect locally.
>   --pod                                           The name of the pod to redirect locally.
>   --service                                       The name of the service to redirect locally.
>   --namespace                                     The namespace containing the resource to redirect locally.
>   --context                                       The kubeconfig context to use. To use this flag, --namespace must also be passed.
>   --local-port                                    Application's port number when running locally.
>   --routing                                       The header value to route on.
>   --use-kubernetes-service-environment-variables  An option to run the connect command when cluster service environment variables are used. To learn more, visit https://aka.ms/use-k8s-svc-env-vars
>   -y|--yes                                        Overrides prompt for confirmation.
> 
> Additional Arguments
>   Any arguments passed following the argument separator "--" will be interpreted as a command to execute in the Bridge
>   environment. This can be used to connect to Kubernetes and start a service in one command line invocation.
>   Example:
>   dsc connect --service my-service --local-port 8000 --routing me -y -- dotnet watch --project MyService.API

The base implementation of `GetHelpText` is doing a lot of heavy lifting... I'm not convinced it is worth it to override it at this point.

elenavillamil (approved):
Change looks good. Thank you for your contribution!

elenavillamil (approved):
Thank you for your contribution! I removed and tested this locally. Change looks good.

PR #27, opened 2022-08-26
Binaries analysis using Binskim tool for SDL

hsubramanianaks:
Hello All, 
1. This PR is a github action to scan generated binaries for bridge CLI using binskim tool and fail if any vulnerabilities is found.
2. This is required as part of SDL (Secure Development Lifecycle) process.
3. Here is the work items linked to it - https://msazure.visualstudio.com/CloudNativeCompute/_workitems/edit/15304625

Thank you for looking into this PR.

### DIFF
--- a/.github/workflows/binaries-analysis.yml
+++ b/.github/workflows/binaries-analysis.yml
@@ -40,5 +40,6 @@ jobs:
         inlineScript: |
           Install-Package Microsoft.CodeAnalysis.BinSkim
           ${{ github.workspace }}\binskim.exe analyze *.exe *.dll --recurse
+        azPSVersion: '3.1.0'
 
     
--- a/.github/workflows/binaries-analysis.yml
+++ b/.github/workflows/binaries-analysis.yml
@@ -2,7 +2,7 @@ name: "Binaries Analysis"
 
 on:
   push:
-    branches: [ "*" ]
+    branches: [ "main" ]
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [ "main" ]
--- a/.github/workflows/binaries-analysis.yml
+++ b/.github/workflows/binaries-analysis.yml
@@ -40,5 +40,3 @@ jobs:
         inlineScript: |
           ${{ github.workspace }}\Microsoft.CodeAnalysis.BinSkim.1.9.5\tools\netcoreapp3.1\win-x64\BinSkim.exe analyze *.exe *.dll --recurse
         azPSVersion: '3.1.0'
-
-    
--- a/.github/workflows/binaries-analysis.yml
+++ b/.github/workflows/binaries-analysis.yml
@@ -12,7 +12,7 @@ on:
 jobs:
   analyze:
     name: BinSkim Binary Analyze
-    runs-on: ubuntu-latest
+    runs-on: windows-latest
     strategy:
       matrix:
         mode: [osx-x64,linux-x64,win-x64]
@@ -35,6 +35,6 @@ jobs:
     - name: Build endpointmanagerlauncher
       run: dotnet publish src\EndpointManagerLauncher\endpointmanagerlauncher.csproj -r win-x64 -c Release --no-restore
     - name: Run BinSkim analysis
-      run: binskim analyze *.exe *.dll --recurse
+      run: binskim.exe analyze *.exe *.dll --recurse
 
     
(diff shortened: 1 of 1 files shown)


.github/workflows/binaries-analysis.yml
```diff
@@ -0,0 +1,44 @@
+name: "Binaries Analysis"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "main" ]
+  schedule:
+    - cron: '21 17 * * 0'
```
Tatsinnit, line 10: 💡☕️ Idea share: Is this very specific for some reasons? i.e. So `* * * * *` means `every minute` of `every hour` of `every day` of `every month` and `every day` of the week. Just checking, 21st min of every 17 hours  or every day to very month and only run on Sundays. Thank you for this.

.github/workflows/binaries-analysis.yml
```diff
@@ -0,0 +1,44 @@
...
+      run: dotnet publish src\dsc\dsc.csproj -c Release -r ${{ matrix.mode }} --no-restore --self-contained true --verbosity normal
+    - name: Build endpointmanagerlauncher
+      run: dotnet publish src\EndpointManagerLauncher\endpointmanagerlauncher.csproj -r win-x64 -c Release --no-restore
+    - name: Install and Run BinSkim analysis
+      uses: Azure/powershell@v1
+      with: 
+        inlineScript: |
+          ${{ github.workspace }}\Microsoft.CodeAnalysis.BinSkim.1.9.5\tools\netcoreapp3.1\win-x64\BinSkim.exe analyze *.exe *.dll --recurse
+        azPSVersion: '3.1.0'
+
```
Tatsinnit, line 43: :💡- nit: few extra carriage returns in the end of this file.

.github/workflows/binaries-analysis.yml
```diff
@@ -0,0 +1,44 @@
...
+    - name: Checkout repository
+      uses: actions/checkout@v3
+    - name: Setup NuGet.exe for use with actions
+      uses: NuGet/setup-nuget@v1.0.6
+    - name: Get BinSkim package from nuget
+      run: nuget install Microsoft.CodeAnalysis.BinSkim -Version 1.9.5 -OutputDirectory ${{ github.workspace }}
+    - name: Restore dependencies using Nuget
+      run: nuget restore src\client.sln -Verbosity Normal -NonInteractive -ConfigFile src/nuget.config
+    - name: dotnet Publish for ${{ matrix.mode }}
+      run: dotnet publish src\dsc\dsc.csproj -c Release -r ${{ matrix.mode }} --no-restore --self-contained true --verbosity normal
```
Tatsinnit, line 34: ☕️ we are using `dotnet` here, but I can't see the botnet action used before its use, how is this workflow getting the `dotnet`, might be we want something like this? https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-net#using-the-net-starter-workflow

Tatsinnit (commented):
🙏☕️💡 Thank you so much for adding me for the review, added few comments rest looks cool, mostly comments to understand the workflow.

.github/workflows/binaries-analysis.yml
hsubramanianaks: removed the extra spaces, thank you,

.github/workflows/binaries-analysis.yml
hsubramanianaks: since I am using windows-latest runner, it comes with dot net pre-installed, please refer here - https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md

.github/workflows/binaries-analysis.yml
hsubramanianaks: Nope, just followed the default settings for Cron job, it runs every Sunday at 5:21. If you want to change this, please let me know or do you think this schedule cron is not necessary please let me know. Thank you.

Tatsinnit: ❤️🙏 thank you for the reply, I ask because see line 5 - we do this binary check for pull-request, push to main branch. So, running on Sunday is just one time thing which is well thought then sounds good to me, please do resolve this conversation as I have approved it. ☕️😊 thanks heaps.

hsubramanianaks: Thank you, closing this conversation.

PR #70, opened 2022-10-07
Ingress port issue

hsubramanianaks:
This PR fixes the ingress port issue and adds extra logger for envoy config and logger statements to find out what port is being used by ingress/envoy. 

**_Testing Done ?_** 

- Yes , Manual testing, please refer screenshots below from routing manager logs
Scenario 1: when port name is provided in the ingress for the service 

image

Scenario 2: when port number is provided in the ingress for the service

image

### DIFF
--- a/src/routingmanager/RoutingManagerApp.cs
+++ b/src/routingmanager/RoutingManagerApp.cs
@@ -327,8 +327,9 @@ namespace Microsoft.BridgeToKubernetes.RoutingManager
                     {
                         var serviceName = path?.Backend?.Service?.Name;
                         var servicePort = path?.Backend?.Service?.Port?.Name;
-                        if (!string.IsNullOrWhiteSpace(servicePort)) {
-                            servicePort = path?.Backend?.Service?.Port?.Number;
+                        if (string.IsNullOrEmpty(servicePort)) {
+                            servicePort = path?.Backend?.Service?.Port?.Number.ToString();
+                            _log.Info("service port name not found for ingress, port number is {0}", servicePort);
                         }
                         if (!string.IsNullOrWhiteSpace(serviceName))
                         {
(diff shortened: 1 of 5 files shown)


bfoley13 (dismissed):
Clean fix, approving with condition tests come later!

Tatsinnit (approved):
👍☕️ thank you for sharing this. 🥷

src/routingmanager/RoutingManagerApp.cs
```diff
@@ -327,8 +327,9 @@ private async Task<(IEnumerable<V1Ingress>, IEnumerable<V1Service>, IEnumerable<
                     {
                         var serviceName = path?.Backend?.Service?.Name;
                         var servicePort = path?.Backend?.Service?.Port?.Name;
-                        if (!string.IsNullOrWhiteSpace(servicePort)) {
-                            servicePort = path?.Backend?.Service?.Port?.Number;
+                        if (string.IsNullOrEmpty(servicePort)) {
+                            servicePort = path?.Backend?.Service?.Port?.Number.ToString();
```
Tatsinnit, line 331: 💡 Just fyi: there will be an issue with this conversion:  If ever magically the name type changes. We could just lock it to `string` could do it like: https://dotnetfiddle.net/srx9kM  

This is so legacy out that for safe refactoring I guess you are the best one to make this choice, just sharing my 2 cents from build break. 

`Cannot implicitly convert type 'int?' to 'string'` -  because of var usage. Wonder just turn it into ternary operation on the check and , are it one liner. 

`string var = someinitalization` - then the number will be treated as string which we are already doing which means any abuse will be blocked. Again all this is just knowing types.

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6791510&view=logs&j=a4f1cc8a-0ccb-5e90-36f4-f06f41d0ce59&t=fc7d1f56-def7-57c7-e3c0-abd90a9d9b80

hsubramanianaks: @Tatsinnit after reading your comments again, I see your point. I will change the variable from var to string so that it is typed correctly. :)

Tatsinnit (approved):

bfoley13 (approved):

Issue #35, opened 2022-09-01
Setting useKubernetesServiceEnvironmentVariables to true causes crash


tomwganem:
This is a copy of https://github.com/microsoft/mindaro/issues/236 which I ran into today while trying to use VSCode and bridge-to-kubernetes on a remote linux machine.

#### Describe the bug
When I set useKubernetesServiceEnvironmentVariables to 'true' (defaults to 'false') in the bridge-to-kubernetes.resource definition in tasks.json and launch the debugger the startup crashes with the following error:

```
Failed to establish a connection. Error: Port 55049 is already in use on your machine. Please free this port and try again. Use 'netstat -ano' command to find which program is using this port.
```

#### To Reproduce
After configuring for local tunneling, edit the bridge-to-kubernetes.resource task in tasks.json and set the useKubernetesServiceEnvironmentVariables entry to 'true'. Start a debug session and watch the terminal output to see the crash, followed by a popup asking if you want to abort.

#### Expected behavior
The service should launch locally and run with environment variables set based on those defined in the Kubernetes cluster.

#### Logs
[...]

PR #78, opened 2022-10-19
Fixing port assignment for headless service

elenavillamil:
Fix for https://github.com/Azure/Bridge-To-Kubernetes/issues/35

Summary: when headless services are used and useKubernetesServiceEnvironmentVariables is set, bridge assigns same port to multiple services. This leads to bridge failing later on when setting up forwarding. The reason why the multiple assignment happens is because we were using same ports object to initialize multiple endpoints. When the local port assignment happens later on, assigning port to one endpoint updates the endpoint of all of them (since ports object shares same place in memory). 

Fix: use new created ports object for each endpoint.

Verification: UTs have been added and end to end verification is in progress.

### DIFF
--- a/src/library/Connect/WorkloadInformationProvider.cs
+++ b/src/library/Connect/WorkloadInformationProvider.cs
@@ -440,8 +440,6 @@ namespace Microsoft.BridgeToKubernetes.Library.Connect
                 var isInWorkloadNamespace = StringComparer.OrdinalIgnoreCase.Equals(endpoint.Metadata.Namespace(), workloadNamespace);
                 foreach (var subset in endpoint.Subsets)
                 {
-                    ports = subset.Ports?.Where(port => this._IsSupportedProtocol(port.Protocol, endpoint.Metadata.Name)).Select(p => new PortPair(remotePort: p.Port)).ToArray() ?? new PortPair[] { };
-
                     // Next, endpoint addresses can also specify their own host names. We add these as well
                     var addresses = subset.Addresses?.Where(a => !string.IsNullOrWhiteSpace(a?.Hostname));
                     if (addresses == null || !addresses.Any())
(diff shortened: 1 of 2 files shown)


Tatsinnit (dismissed):
👍🎊 Thank you so adding me to the review, looks good to me.

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -456,7 +454,7 @@ private int GetServicePortValue(V1ServicePort servicePort, V1Container container
                             DnsName = isInWorkloadNamespace ?
                                         $"{address.Hostname}.{endpoint.Metadata.Name}" :
                                         $"{address.Hostname}.{endpoint.Metadata.Name}.{endpoint.Metadata.Namespace()}",
-                            Ports = ports,
+                            Ports = subset.Ports?.Where(port => this._IsSupportedProtocol(port.Protocol, endpoint.Metadata.Name)).Select(p => new PortPair(remotePort: p.Port)).ToArray() ?? new PortPair[] { },
```
hsubramanianaks, line 457: Hi Elena, thanks for including me. so now instead of setting ports for each subset of endpoints, you have moved it to for each of list of address for endpoints ? I see it is same statement moved from the top, wondering if this worked ?

elenavillamil: so the original assignment was correct in terms of content, only problem was because it was same object it shared same memory address (this is a C# thing of shallow copy vs deep copy all object except structs in c# are by reference). By moving it here now content is still original content it was using before but each endpoint gets its own object (new keyword for each assignment) so later on changing the value of a port on any endpoint wont change the value of the port in all endpoints.

hsubramanianaks (approved):
LGTM we need to make sure Joyer's test this well before moving to production.

Issue #84, opened 2022-10-30
Installation script - Windows scenario: Reverify the aspnetcore-runtime-3.1 existence for choco.


Tatsinnit:
**Describe the bug**

For the new installation script for `windows` O/S the `choco` packager tries to install `aspnetcore-runtime-3.1` which is not there

The correct package it seems is following: https://community.chocolatey.org/packages/dotnetcore-3.1-aspnetruntime

* Code ref: https://github.com/Azure/Bridge-To-Kubernetes/blob/main/scripts/install.sh#L128

**Mention the platform you are using**
- Windows O/S 
```
$ cmd /c ver
Microsoft Windows [Version 10.0.22621.674]
(c) Microsoft Corporation. All rights reserved.
```

**To Reproduce**
1. wsl or git bash and then run:
2. `curl -fsSL https://raw.githubusercontent.com/Azure/Bridge-To-Kubernetes/main/scripts/install.sh | bash`

**Expected behavior**
The script should run successfully

**Screenshots**

![image](https://user-images.githubusercontent.com/6233295/198899896-ad0a0812-1a9b-4f1a-91f7-3b9aae8cef8f.png)

![image](https://user-images.githubusercontent.com/6233295/198899913-0613cc2c-e8e4-44e1-88c0-2638e4349877.png)

PR #85, opened 2022-10-30
Fix choco endpoint.

Tatsinnit:
I found that the choco endpoitn was incorrect, so not sure if this got checked in but fyi @hsubramanianaks for double chekcing here is the fix PR and the choco endpoint: https://community.chocolatey.org/packages/dotnetcore-3.1-aspnetruntime

More details: #84 

Thanks,

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -129,8 +129,10 @@ install_tool() {
             if [[ $OSTYPE == "darwin"* ]]; then
                 $PACKAGER tap isen-ng/dotnet-sdk-versions
                 install_with_sudo dotnet-sdk3-1-400 --cask
-            else 
+            else if [[ $OSTYPE == "msys"* ]]; then
                 install_with_sudo dotnetcore-3.1-aspnetruntime
+            else 
+                install_with_sudo aspnetcore-runtime-3.1
             fi
             ;;
         jq)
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -130,7 +130,7 @@ install_tool() {
                 $PACKAGER tap isen-ng/dotnet-sdk-versions
                 install_with_sudo dotnet-sdk3-1-400 --cask
             else 
-                install_with_sudo aspnetcore-runtime-3.1
+                install_with_sudo dotnetcore-3.1-aspnetruntime
```
hsubramanianaks, line 133: We need add a condition for msys which is git bash or else wsl or Linux will break. Thank you for catching this.

Tatsinnit: Yeah, sounds like this should handle it?

Also, pkm choco reference - https://community.chocolatey.org/packages/dotnetcore-3.1-aspnetruntime 

Please do recheck if previous was ever working. Or is this the case that `aspnetcore-runtime-3.1` works for the `ubuntu` and not for the windows?

```*.sh
install_with_sudo() {
    if [[ $OSTYPE == "linux"* ]]; then
        sudo $PACKAGER install $1 -y
    else
        $PACKAGER install $2 $1
    fi
}
```

hsubramanianaks:
Closing this PR, we will work on creating a new one from forked repo.

PR #88, opened 2022-11-02
Fix Managed Identity when with useKubernetesServiceEnvironmentVariables

elenavillamil:
Background: Managed Identity feature of bridge to kubernetes was not working when user ran under useKubernetesServiceEnvironmentVariables. Bridge adds MSI_ENDPOINT environment variable as a DNS name for the "fake" manage identity service bridge starts (http://managedidentityforbridgetokubernetes/metadata/identity/oauth2/token). When user runs bridge as useKubernetesServiceEnvironmentVariables we dont have access to add entry in host files and thus when MSI library tried to hit that endpoint there was a DNS resolution error.  

With this fix we make sure that when user runs as useKubernetesServiceEnvironmentVariables we add MSI_ENDPOINT using IP:Port instead of DNS name. Something like http://127.0.0.1:5050/metadata/identity/oauth2/token

This PR also adds UT coverage for that function. 

Integration tests done as well.

### DIFF
--- a/src/library/Connect/LocalEnvironmentManager.cs
+++ b/src/library/Connect/LocalEnvironmentManager.cs
@@ -299,7 +299,7 @@ namespace Microsoft.BridgeToKubernetes.Library.Connect
             using (var perfLogger = _log.StartPerformanceLogger(Events.LocalEnvironmentManager.AreaName, Events.LocalEnvironmentManager.Operations.GetLocalEnvironment))
             {
                 await this._LoadAdditionalServiceEnvAsync(remoteAgentLocalPort, workloadInfo, localProcessConfig, cancellationToken);
-                var result = this._CreateEnvVariablesForK8s(workloadInfo);
+                var result = this.CreateEnvVariablesForK8s(workloadInfo);
                 perfLogger.SetSucceeded();
                 perfLogger.SetProperty("EnvVarCount", result.Count);
                 return result;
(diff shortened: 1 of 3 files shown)


src/library/Connect/LocalEnvironmentManager.cs
```diff
@@ -415,7 +415,7 @@ private void _ReportProgress(EventLevel eventLevel, string message, params objec
         /// <summary>
         /// Re-create the Kubernetes service related environment variables.
         /// </summary>
-        private IDictionary<string, string> _CreateEnvVariablesForK8s(WorkloadInfo workloadInfo)
+        public IDictionary<string, string> CreateEnvVariablesForK8s(WorkloadInfo workloadInfo)
```
elenavillamil, line 418: Changing this method to public so I could add UT coverage for it. Feel free to point out if there is better way to achieve this.

src/library/Connect/LocalEnvironmentManager.cs
```diff
@@ -468,6 +468,12 @@ private void _AddKubernetesServiceEnv(IDictionary<string, string> envVariables,
             envVariables[$"{serviceName}_SERVICE_PORT"] = port.ToString();
             envVariables[$"{serviceName}_PORT_{port}_{protocolUpper}_ADDR"] = host;
             envVariables[$"{serviceName}_SERVICE_HOST"] = host;
+
+            // if this is managed identity with useKubernetesServiceEnvironmentVariables set to true we have to update ms endpoint variable
+            // from dns name to ip:port 
+            if (this._useKubernetesServiceEnvironmentVariables && StringComparer.OrdinalIgnoreCase.Equals(serviceName, Common.Constants.ManagedIdentity.TargetServiceNameOnLocalMachine)) {
+                envVariables[ManagedIdentity.MSI_ENDPOINT_EnvironmentVariable] = "http://" + host + ":" + port + "/metadata/identity/oauth2/token";
```
Tatsinnit, line 475: 💡☕️ Is there anything like “https” for this context? If that’s the case might be enum.

Tatsinnit (approved):
🤞☕️❤️🎉🥷 thanks for adding me to this review! Looks good to me! I think you and Hari might know more context etc.

hsubramanianaks (approved):
LGTM

src/library/Connect/LocalEnvironmentManager.cs
elenavillamil: Nope, only http. To my knowledge there are no plans to make this or add https.

PR #93, opened 2022-11-09
Address Managed Identity workflow for Go library

elenavillamil:
Current Managed Identity workflow works with C# MI library, but it was erroring out when MI was being used from Go library. Request was being updated per below, **this** was being added in the wrong place. This PR simplifies and fixes the function that updates the request before sending to real manage identity. 

2022-11-04T00:47:46.2175539Z | RemoteAgent | TRACE | ManagedIdentity :  Original data : GET /metadata/identity/oauth2/token?api-version=2017-09-01&mi_res_id=%2Fsubscriptions%2F4be8920b-2978-43d7-ab14-04d8549c1d05%2FresourceGroups%2FAKSE2EInfra%2Fproviders%2FMicrosoft.ManagedIdentity%2FuserAssignedIdentities%2\nAccept-Encoding: gzip (amd64-linux) go-autorest/adal/v1.0.0.1
2022-11-04T00:47:46.2178054Z | RemoteAgent | TRACE | ManagedIdentity :  Sending data : **Metadata: true\n\n**GET /metadata/identity/oauth2/token?api-version=2018-02-01&mi_res_id=%2Fsubscriptions%2F4be8920b-2978-43d7-ab14-04d8549c1d05%2FresourceGroups%2FAKSE2EInfra%2Fproviders%2FMicrosoft.ManagedIdentity%2FuserAss\nAccept-Encoding: gzip (amd64-linux) go-autorest/adal/v1.0.0ure.com%2F HTTP/1.1

This change has been tested with Go library and with C# library for backwards compatibility.

### DIFF
--- a/src/devhostagent/PortForward/ServicePortForwardConnector.cs
+++ b/src/devhostagent/PortForward/ServicePortForwardConnector.cs
@@ -178,6 +178,7 @@ namespace Microsoft.BridgeToKubernetes.DevHostAgent.PortForward
         /// <param name="data"></param>
         /// <remarks>
         /// input data should look like below:
+        /// GET /metadata/identity/oauth2/token?api-version=2017-09-01&mi_res_id=%2Fsubscriptions%2F4be8920b-2978-43d7-ab14-04d8549c1d05%2FresourceGroups%2FAKSE2EInfra%2Fproviders%2FMicrosoft.ManagedIdentity%2FuserAssignedIdentities%2Fakse2ehcp&resource=https%3A%2F%2Fstorage.azure.com%2F
         /// GET /metadata/identity/oauth2/token?api-version=2017-09-01&resource=https%3A%2F%2Fstorage.azure.com&clientid=<guid> HTTP/1.1
         /// \nHost: managedidentity
         /// \nsecret: placeholder
@@ -196,25 +197,35 @@ namespace Microsoft.BridgeToKubernetes.DevHostAgent.PortForward
         /// </remarks>
         private byte[] GetModifiedBytesForManagedIdentity(byte[] data)
         {
-            const string headerToInsert = "Metadata: true\n";
-            var stringContent = Encoding.Default.GetString(data.ToList().ToArray());
+            const string headerToInsert = "Metadata: true";
+            var stringContent = Encoding.Default.GetString(data);
+            // This log may not work as expected depending on the character on the string, adding print line inside 
+            // the for lop below to avoid wrong information while debugging
             _log.Verbose($"{loggingPrefix} Original data : {stringContent}");
             stringContent = stringContent.Replace("2017-09-01", "2018-02-01").Replace("clientid", "client_id");
             var lines = stringContent.Split(new char[] { '\n' }).ToList();
             int index = -1;
             for (int i = 0; i < lines.Count; i++)
             {
-                if (lines[i].Contains($"secret: {Common.Constants.ManagedIdentity.SecretValue}"))
+                _log.Verbose($"{loggingPrefix} Printing line : {lines[i]}");
+                if (lines[i].Contains($"secret: {Common.Constants.ManagedIdentity.SecretValue}", StringComparison.OrdinalIgnoreCase))
                 {
                     index = i;
                     break;
                 }
             }
-            lines.Insert(index + 1, headerToInsert);
+            if (index != -1) {
+                lines.Insert(index + 1, headerToInsert);
+            }
+            else {
+                lines.Append(headerToInsert);
+            }
             var modifiedRequestString = string.Join('\n', lines);
 
             _log.Verbose($"{loggingPrefix} Sending data : {modifiedRequestString}");
-            return Encoding.ASCII.GetBytes(modifiedRequestString);
+
+            return Encoding.Default.GetBytes(modifiedRequestString);
+
         }
 
         #endregion private members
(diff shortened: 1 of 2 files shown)


src/devhostagent/PortForward/ServicePortForwardConnector.cs
```diff
@@ -196,25 +197,35 @@ private async Task HookupStreamData(StreamHookData s)
...
-                if (lines[i].Contains($"secret: {Common.Constants.ManagedIdentity.SecretValue}"))
+                _log.Verbose($"{loggingPrefix} Printing line : {lines[i]}");
+                if (lines[i].Contains($"secret: {Common.Constants.ManagedIdentity.SecretValue}", StringComparison.OrdinalIgnoreCase))
                 {
                     index = i;
                     break;
                 }
             }
-            lines.Insert(index + 1, headerToInsert);
+            if (index != -1) {
```
Tatsinnit, line 217: 🥷☕️ Please note: This is **interesting**, why have we changed this? Reason I ask is - previously is was `lines.Insert(index + 1, headerToInsert);` which will take care of all the cases, which means if its `-1` then place the value on the `0th` index because `-1 + 1 == 0` otherwise for any `non-0th` index use the index value. Line-207 - guarantee that first inter will be `-1` 

Why are we changing this to specific if-and-else please?

src/devhostagent/PortForward/ServicePortForwardConnector.cs
```diff
@@ -196,25 +197,35 @@ private async Task HookupStreamData(StreamHookData s)
         /// </remarks>
         private byte[] GetModifiedBytesForManagedIdentity(byte[] data)
         {
-            const string headerToInsert = "Metadata: true\n";
-            var stringContent = Encoding.Default.GetString(data.ToList().ToArray());
+            const string headerToInsert = "Metadata: true";
+            var stringContent = Encoding.Default.GetString(data);
```
Tatsinnit, line 201: 💡 Please note: I am not entirely sure of this change, but adding some thoughts: `GetString()` take byte array we are changing the data contract from what it used to be to `data` so is this contract changed recently?

src/devhostagent/PortForward/ServicePortForwardConnector.cs
```diff
@@ -196,25 +197,35 @@ private async Task HookupStreamData(StreamHookData s)
...
+            }
+            else {
+                lines.Append(headerToInsert);
+            }
             var modifiedRequestString = string.Join('\n', lines);
 
             _log.Verbose($"{loggingPrefix} Sending data : {modifiedRequestString}");
-            return Encoding.ASCII.GetBytes(modifiedRequestString);
+
+            return Encoding.Default.GetBytes(modifiedRequestString);
```
Tatsinnit, line 227: 💡☕️ Since I don't know much of background, I will add this thought that - Please note **It is not recommended** to use `Encoding.Default`

* https://learn.microsoft.com/en-us/dotnet/api/system.text.encoding.default?redirectedfrom=MSDN&view=net-7.0#remarks

> Different computers can use different encodings as the default, and the default encoding can even change on a single computer. Therefore, data streamed from one computer to another or even retrieved at different times on the same computer might be translated incorrectly. In addition, the encoding returned by the Default property uses best-fit fallback to map unsupported characters to characters supported by the code page. For these two reasons, using the default encoding is generally not recommended. To ensure that encoded bytes are decoded properly, your application should use a Unicode encoding, such as UTF8Encoding or UnicodeEncoding, with a preamble. Another option is to use a higher-level protocol to ensure that the same format is used for encoding and decoding.

Tatsinnit (commented):
💡❤️☕️🙏 Thank you so much for adding me to this review, I have added few comments for this PR, Hope it helps. Thanks heaps.

src/devhostagent/PortForward/ServicePortForwardConnector.cs
elenavillamil: This is an HTTP request, we are changing the raw request, if we insert Metadata:True as first line (aka index 0) it will invalidate the request since the first line should start with request type (GET, POST, etc). If you check the description of the PR I have added the logs of what it looked like when I started debugging, it changed my request from GET /metadata/identity/oauth2/token?api-versio.... to Metadata: true\n\nGET /metadata/identity/oauth2/token?api-versio... 

Basically original code had an assumption that the secret header was always present, but if it isnt we should append the metadata header at the end not at the begining. Hope this helps.

src/devhostagent/PortForward/ServicePortForwardConnector.cs
elenavillamil: this is my attempt at simplifying this code. I have tested with both ways and there is no regression. data is a byte array. I do not see the need to go from bytes array to list and then to array again if the GetString can work with data directly.

src/devhostagent/PortForward/ServicePortForwardConnector.cs
elenavillamil: this is a good point. Let me look into it. I know ASCII was breaking Go, but let me see if I can UTF8.

ShantingLiu (approved):

hsubramanianaks (approved):
Looks good, this will work with other libraries ex: Java? If an app is using Java library with Managed Identity, I know we don't have a sample application to test but we want to keep eye on those. Please correct me if I am wrong :)

PR #48, opened 2022-09-11
Update to net 6.0 

pkoelbl:
The purpose of this is to update to the current LTS for .net(6.0). 

It is also to enable arm64 support for M1 Macs.

PR #106, opened 2022-11-17
Dotnet 6 Upgrade

hsubramanianaks:
Thanks @elenavillamil heaps for suggesting the fix. Thank you @pkoelbl for creating this PR. 

This PR from the main [PR](https://github.com/Azure/Bridge-To-Kubernetes/pull/48) plus the fix for control port stop issue. 

Thanks everyone who have checked this PR.


Tatsinnit:
💡 ❤️ Thankyou for adding me to this review! A lot of this is going to be a huge `/hallpass` from me because of the sheer amount of files. :) 

I would recommend if we can tag our amazing test folks here as well and if they can please run this before we merge this. Worth getting more eye, and thoughts on this idea. LGTM (some CodeQL breaks but thats normal for this size of this PR and visible)

Many File

PR #129, opened 2022-12-12
Update devhostagent port to be different from GRPC unofficial port

elenavillamil:
DevHostAgent was using GRPC unofficial port for connection with bridge in local cluster. Update it to different port to avoid conflicts with GRPC services.

Adding warning log to alert user if their service port is ever the same as dev host agent port. This port is shared across DevHostAgent, CLI and restoration job. Updating on the fly based on user's service port adds complexity that does not seems needed at the moment. We can re-evaluate later if this becomes an issue.


bfoley13 (dismissed):
lgtm!

src/common/DevHostAgent/DevHostConstants.cs
```diff
@@ -9,7 +9,7 @@ internal static class DevHostConstants
     {
         internal static class DevHostAgent
         {
-            public const int Port = 50051;
+            public const int Port = 50052;
```
Tatsinnit, line 12: 💡 idea / thoughts: This port is used with some specific use-case or is it just increment by 1 because GRPC services is using the other `50051` port? If so can just an idea that can we get something fixated for B2K cli and document it? cc: @hsubramanianaks as fyi and thoughts.

Tatsinnit (commented):
❤️☕️💡 Thank you so much for adding me to this review, I've added couple of things as what I noticed for some clarity, thanks heaps, @hsubramanianaks for his eyes as well. thank you.

src/library/Connect/LocalEnvironmentManager.cs
```diff
@@ -226,7 +226,11 @@ CancellationToken cancellationToken
                 {
                     this._reversePortForwardManager.Start(pfInfo);
                     this._ReportProgress(Resources.ContainerPortAvailableFormat, pfInfo.Port, pfInfo.LocalPort);
+                    if (pfInfo.Port == DevHostConstants.DevHostAgent.Port) {
+                        _log.Warning($"Service to be debug uses same port '{pfInfo.Port}' as Bridge DevHostAgent service does. This can cause conflict. It's recommended to update service to use different port.");
+                    }
                 }
+                perfLogger.SetSucceeded();
```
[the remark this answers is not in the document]
elenavillamil: If you look at the library logs for a run you will see below log for LocalEnvironmentManager-StarReversePortForwarding: 
2022-12-08T21:38:51.5539847Z | Library | TRACE | Event: **LocalEnvironmentManager-StartReversePortForwarding** <json>{"properties":{"**result":"Failed**"},"metrics":{"durationInMs":8.0}}</json>

While if you look at log for LocalEnvironmentManager-StartServicePortForwardings you will see below log
2022-12-12T19:19:09.2574798Z | Library | TRACE | Event: LocalEnvironmentManager-StartServicePortForwardings <json>{"properties":{"result":"Succeeded"},"metrics":{"durationInMs":10.0}}</json>

Note it is different to the one that comes from ConnectMaangementClient.
2022-12-12T19:19:09.2621246Z | Library | TRACE | Event: ConnectManagementClient-StartServicePortForwardings <json>{"properties":{"result":"Succeeded"},"metrics":{"durationInMs":16.0}}</json>

Essentially, everytime you start a PerfLogger it will log that log line when it was out of scope (this is why there is 3 log lines, one for each start and one for the actions that starts all the port forwarding), and it initial state/result if not set is False.

src/library/Connect/LocalEnvironmentManager.cs
```diff
@@ -226,7 +226,11 @@ CancellationToken cancellationToken
                 {
                     this._reversePortForwardManager.Start(pfInfo);
                     this._ReportProgress(Resources.ContainerPortAvailableFormat, pfInfo.Port, pfInfo.LocalPort);
+                    if (pfInfo.Port == DevHostConstants.DevHostAgent.Port) {
+                        _log.Warning($"Service to be debug uses same port '{pfInfo.Port}' as Bridge DevHostAgent service does. This can cause conflict. It's recommended to update service to use different port.");
```
hsubramanianaks, line 230: nit Service to be debugged uses same port, can you correct it please ?

src/common/DevHostAgent/DevHostConstants.cs
elenavillamil: Screenshot 2022-12-13 at 10 38 17 AM

Tatsinnit: Sounds good, cool, lets fyi @qpetraroia  for the documentation update, or if we its something small and we want it to be part of this PR? , also, thinking out loud for idea but I hope this will not introduce any breaking change for folks who have dependency on this this for port which was used by B2K. (just in case if its day-0 port.)  

Please feel free to resolve this conversation if you are happy. thanks heaps. ☕️☕️👍

Tatsinnit (dismissed):
🎉❤️☕️

bfoley13 (approved):

hamzawajahat (commented):
1 to 4 remove

Issue #56, opened 2022-09-16
Multiple "proxy" pods running, blocking a new debug session


Eneuman:
Using Visual Studio Professional 2021 with latest extension.

If you stop your debug session and then start it while your previous pod is being restored by the "restore" pod, you sometimes ends up with two "proxy" pods running, both theese pod keeps craching and the only way to fix this is to deploy you original pod from scratch.

I think I have figured out what is happening and how to fix this.
The problem seems to be that sometimes a "proxy" pod is created that tries to proxy the already existing proxy pod.

How to fix this:
If you try to start debugging when there are two pods running, you will get a error message from the extension, telling you that  "... is currently backed by two pods. Please try again later".

If this code were to look at the tags set on thoose pods, it would see that both pods are "proxy" pods, created by B2K. It could then just remove both theese pods and restore the original pod, before trying to setup a new debug session.

It's easy to reproduce so let me know if you need any logs.

elenavillamil:
Hi @Eneuman, before starting a new Bridge connection you should always wait for previous connection to be fully restored. If you start a new connection before previous one is properly restored. There are several things that go wrong when a new session is connection before previous one is cleaned up. For bridge to try an clean up that state is too complicated, but what we can do is add a check when bridge is connecting and if it sees <your service>-restore pod is present then show error to the user and prevent connection.

Eneuman:
@elenavillamil A check would be great. Right now i'm using k9s to see when the restore pod has finished it's job. But sometimes I forget to check and then I have to redeploy the pod from the devops pipeline, witch takes time.

Eneuman:
> For bridge to try an clean up that state is too complicated, but what we can do is add a check when bridge is connecting and if it sees -restore pod is present then show error to the user and prevent connection.

After having a second look at this, maby using kubernets "roullout undo" could be a nice way of handling this.

Running this command `kubectl rollout history deploy myservice` shows all changes made to the service.

```
REVISION  CHANGE-CAUSE
810       <none>
811       <none>
813       <none>
815       <none>
817       <none>
819       <none>
821       <none>
823       <none>
825       <none>
827       <none>
828       <none>
```


If the B2K deployment would set the "CHANGE-CAUSE" to "B2K", then it could loop through the revisions to find the latest not marked with B2K and then do a `kubectl rollout undo deploy myservice--to-revision=X`
and after that, delete and restore pod.

Doing this manually seems to work.

elenavillamil:
[...]

PR #136, opened 2022-12-27
Show warning if on connection service has restore pod

elenavillamil:
This change is related to issue: https://github.com/Azure/Bridge-To-Kubernetes/issues/56. The goal is to improve usability when user is trying to connect bridge but previous session has not been restore yet. After this change, if restore pod is present for the target service we wont allow a new connection and let user know. Before


Marcus-Hines (dismissed):
lgtm

src/library/Utilities/RemoteContainerConnectionDetailsResolver.cs
```diff
@@ -49,6 +49,17 @@ internal class RemoteContainerConnectionDetailsResolver
                 return remoteContainerConnectionDetails;
             }
 
+            // Check if restoration pod is present. This is an indication that previous session is still connected or it has not finished restoring yet.
+            var allPods = (await _kubernetesClient.ListPodsInNamespaceAsync(remoteContainerConnectionDetails.NamespaceName, null, cancellationToken))?.Items;
+            var podsRunningRestorationJob = allPods.Where(p => p.Metadata.Name.StartsWith(remoteContainerConnectionDetails.ServiceName + "-restore") &&
```
Tatsinnit, line 54: 💡 Thoughts/A quick note: there are few places where we are using `?` for nullable type check, in case if they are nullable will it be worth logging it? example: `p.Status?.ContainerStatuses?`  (I am not sure of use case hence just idea)

if it is worth to log might be use `HasValue` and log that those value are null etc.

src/library/Utilities/RemoteContainerConnectionDetailsResolver.cs
```diff
@@ -49,6 +49,17 @@ internal class RemoteContainerConnectionDetailsResolver
                 return remoteContainerConnectionDetails;
             }
 
+            // Check if restoration pod is present. This is an indication that previous session is still connected or it has not finished restoring yet.
+            var allPods = (await _kubernetesClient.ListPodsInNamespaceAsync(remoteContainerConnectionDetails.NamespaceName, null, cancellationToken))?.Items;
+            var podsRunningRestorationJob = allPods.Where(p => p.Metadata.Name.StartsWith(remoteContainerConnectionDetails.ServiceName + "-restore") &&
+                                                        (p.Status?.ContainerStatuses?.Where(cs => cs.Image.Contains(ImageProvider.DevHostRestorationJob.Name)).Any() ?? false));
+            if (podsRunningRestorationJob.Count() > 0)
```
Tatsinnit, line 56: 💡☕️🙏 Just idea: Can this ever be null? if so might be check if this function is in utils already.

```.cs
public static bool IsNullOrEmpty(this Array array)
{
    return (array == null || array.Length == 0);
}

**Review by Tatsinnit — APPROVED** [review]
☕️🙏🎊 Thank you so much for adding me to this review. Add3ed few observation mostly around how `null` is handled.

Rest looks good to me. Thanks 🤟

File: src/library/Utilities/RemoteContainerConnectionDetailsResolver.cs (hunk 2 above)
Comment (hsubramanianaks) [reply src/library/Utilities/RemoteContainerConnectionDetailsResolver.cs:56]: FYI - Dotnet framework has a string.IsNullOrEmpty() method which can be used here.

Comment (elenavillamil) [reply src/library/Utilities/RemoteContainerConnectionDetailsResolver.cs:56]: podsRunningRestorationJob will never be null, it will just be an empty list if no matches. Let me know if you were refereing to a variable different than podsRunningRestorationJob.

File: src/library/Utilities/RemoteContainerConnectionDetailsResolver.cs (hunk 1 above)
Comment (elenavillamil) [reply src/library/Utilities/RemoteContainerConnectionDetailsResolver.cs:54]: I think it is safety checks because of how many different service / deployment configurations different users can have. I think if we log every time we have one of those checks the logs would be too noisy and the code more complex (since we would have several more if else levels).

File: src/library/Utilities/RemoteContainerConnectionDetailsResolver.cs (hunk 2 above)
Comment (Tatsinnit) [reply src/library/Utilities/RemoteContainerConnectionDetailsResolver.cs:56]: Cool, sounds good ☕️👍 I mentioned because: if you check in line 53 where new var allpods is defined it uses “?” Nullable type and then at line 54 this var.

Suggestion is - Since you might know the scenario feel feee to use null-forgiving operator where null needs suppressing. 

Please feel free to resolve this conversation if not needed, thanks.

File: src/library/Utilities/RemoteContainerConnectionDetailsResolver.cs (hunk 1 above)
Comment (Eneuman) [reply src/library/Utilities/RemoteContainerConnectionDetailsResolver.cs:54]: It might also be a good idea to check that the pod status is running. If the pod has crashed, starting a new debug session will redeploy the restore pod and fix the problem today.

File: src/library/Utilities/RemoteContainerConnectionDetailsResolver.cs (hunk 2 above)
Comment (elenavillamil) [reply src/library/Utilities/RemoteContainerConnectionDetailsResolver.cs:56]: Thank @Tatsinnit! I understand your comment now. This is a good catch. I did not find function static bool IsNullOrEmpty(this Array array) so I am just doing null check

File: src/library/Utilities/RemoteContainerConnectionDetailsResolver.cs (hunk 1 above)
Comment (elenavillamil) [reply src/library/Utilities/RemoteContainerConnectionDetailsResolver.cs:54]: @Tatsinnit, sorry I misunderstood you. I see what you are saying that if pod is there, then status should never be null. I took usage from existing code previous team had writing, and without knowing historical reason on why they added it to begin with, I will leave it for now.

@Eneuman, this is good point. I am adding that check now.

**Review by Tatsinnit — APPROVED** [review]
☕️💡 LGTM - Thanks heaps for including me in PR.

**Review by Tatsinnit — APPROVED** [review]

**Review by sabbour — APPROVED** [review]

File: src/library/Utilities/RemoteContainerConnectionDetailsResolver.cs [hunk 3]
```diff
@@ -49,6 +49,20 @@ internal class RemoteContainerConnectionDetailsResolver
                 return remoteContainerConnectionDetails;
             }
 
+            // Check if restoration pod is present and in running state. This is an indication that previous session is still connected or it has not finished restoring yet.
+            var allPods = (await _kubernetesClient.ListPodsInNamespaceAsync(remoteContainerConnectionDetails.NamespaceName, null, cancellationToken))?.Items;
```
hsubramanianaks, line 53: Hi, just noticed this, instead of var if we know the type of variable we can instead declare it with the exact type which is safer than var keyword. 
stack over flow reference - https://stackoverflow.com/questions/545616/why-would-var-be-a-bad-thing
```suggestion
 V1PodList v1 = await _kubernetesClient.ListPodsInNamespaceAsync(namespaceName, selectors, cancellationToken);
                IList<V1Pod> items = v1.Items;
                if (items.Any() || items.Count > 0)
                {
// your code here
                }
            var allPods = (await _kubernetesClient.ListPodsInNamespaceAsync(remoteContainerConnectionDetails.NamespaceName, null, cancellationToken))?.Items;

```

hsubramanianaks: @elenavillamil i think you missed my comment. No worries it's just a suggestion Which would increase the quality of code. :)

PR #138, opened 2022-12-28
Update azure-pipelines.yml for Azure Pipelines

hsubramanianaks:
Minor changes to the Azure build pipelines. Thanks for the review.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -63,11 +63,7 @@ jobs:
         if ($branchName -eq "refs/heads/main") {
             Write-Host "##vso[task.setvariable variable=SignTypeToUse]real"
             Write-Host "##vso[task.setvariable variable=TelemetryType]TELEMETRY_PRODUCTION"
-            Write-Host "Done setting the sign type and telemetry key instance for staging branch"
-        }
-
-        else {
-            Write-Host "Not setting the sign type and telemetry key instance for $branchName branch"
+            Write-Host "Done setting the sign type and telemetry key instance for main branch"
         }
 
 
@@ -250,7 +246,7 @@ jobs:
   - task: VSBuild@1
     name: ''
     displayName: Sign Files
-    condition: and(succeeded(), ne(variables['Build.Reason'], 'PullRequest'))
+    condition: and(succeeded(), eq(variables['RunTests'], 'true'))
     enabled: True
     inputs:
       solution: src/client.signproj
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -63,11 +63,7 @@ jobs:
         if ($branchName -eq "refs/heads/main") {
             Write-Host "##vso[task.setvariable variable=SignTypeToUse]real"
             Write-Host "##vso[task.setvariable variable=TelemetryType]TELEMETRY_PRODUCTION"
-            Write-Host "Done setting the sign type and telemetry key instance for staging branch"
```
Tatsinnit, line 66: 💡☕️ For Archival reasons, I believe this is set because we no longer want to keep the staging et. al. anymore?

azure-pipelines.yml
```diff
@@ -250,7 +246,7 @@ jobs:
   - task: VSBuild@1
     name: ''
     displayName: Sign Files
-    condition: and(succeeded(), ne(variables['Build.Reason'], 'PullRequest'))
+    condition: and(succeeded(), eq(variables['RunTests'], 'true'))
```
Tatsinnit, line 249: 💡☕️ Question: Where will this variable be getting set please?

azure-pipelines.yml
```diff
@@ -416,14 +412,6 @@ jobs:
       script: >-
         $ZipHost = "mindaromaster.blob.core.windows.net"
 
-        if ($env:BUILD_SOURCEBRANCH -eq "refs/heads/staging") {
```
Tatsinnit, line 419: 💡☕️ For Archival reasons only please, I believe this is set because we no longer want to keep the staging et. al. anymore?

Tatsinnit (approved):
❤️☕️🙏🎊 Thank you so much for adding me to this review, I have added some archival comments for latter understanding, rest looks cool. 👍

azure-pipelines.yml
hsubramanianaks: Previous team who owned B2K had a different CI/CD process, they would move code from main branch to staging branch and then to production branch. sometimes code to production was cherry picked, but we want cleaner approach. so whatever is in main branch will be released to dev --> staging --> production. Hence checking for staging/production branch checks are removed. I hope this make sense. Thank you.

azure-pipelines.yml
hsubramanianaks: It is set in variables tab , here is screenshot of it. 
image

azure-pipelines.yml
hsubramanianaks: Same reason as previous comment. 
https://github.com/Azure/Bridge-To-Kubernetes/pull/138/files/c7d24b9b9c38cb6925e6eb08ca6aead87b3940e0#r1058975848

azure-pipelines.yml
hsubramanianaks: I moved them from variables tab to variables in the pipeline definition, hopefully that clarifies things.

Issue #89, opened 2022-11-04
How can I configure Bridge to NOT forward certain services' ports?


ptixed:
Hello,

How can I configure Bridge to NOT forward certain services' ports?

I have a smb server on my cluster and I am trying to debug another app that does not depend on it. Bridge tries to allocate port 445 on my local machine and errors out because this port is already in use (I'm on windows, port 445 is in use by default). How can I tell the Bridge to just ignore the smb service and don't perform any forwards?

elenavillamil:
Hi @ptixed, this is not supported today. As a temporary workaround you can change the port your Kubernetes service uses. I have marked this as "Feature Request" and we will investigate as a team if it is possible to add this feature. Thank you.

elenavillamil:
Closing since this has been address with 02/16 release.

PR #144, opened 2023-01-10
Added method to ignore ports

Vidya2606:
The changes are related to - https://github.com/Azure/Bridge-To-Kubernetes/issues/89
In windows machines certain ports are taken up by the OS. When a user tries to runs Kubernetes cluster sometimes the services are allocated ports which are in use by the OS, these attempts fail with error "The ports are already in use".

Solution
- Users should provide ports which should not be used by the service. After creating the service, edit the service by running the following command from CLI
   **Kubectl edit service stats-api**

- Under the metadata add the following annotation, this example ignores ports 445 and 23
   **bridgetokubernetes/ignore-ports: 445,23**

- Save and exit. This lets the service know which ports to be ignored.

The screenshot attached shows how we can let the service know which ports to ignore.
Screenshot 2022-12-29 at 1 16 47 PM


src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -117,6 +120,28 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
             }
         }
 
+        //
+        public List<int> GetPortsToIgnoreFromAnnotations(V1ServiceList serviceNameSpace) {
+            var portsToIgnore = new List<int>();
+            foreach(var item in serviceNameSpace.Items) {
+                if (item.Metadata.Annotations != null) {
```
elenavillamil, line 127: We should probably add check that Metadata is not null as well:
if (item.Metadata != null && item.Metadata.Annotations != null)

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -100,14 +103,14 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
                         reachableEndpoints.Add(new EndpointInfo()
                         {
                             DnsName = externalEndpoint.Name,
-                            Ports = externalEndpoint.Ports.Select(p => new PortPair(remotePort: p)).ToArray(),
+                            Ports = externalEndpoint.Ports.Select(p => new PortPair(remotePort: p)).ToArray(),
```
elenavillamil, line 106: Can we add comment here that we should not take into account ports to ignore here since this code path is handling endpoints user explicitly added for tracking.

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -117,6 +120,28 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
...
+            var portsToIgnore = new List<int>();
+            foreach(var item in serviceNameSpace.Items) {
+                if (item.Metadata.Annotations != null) {
+                    if (item.Metadata.Annotations.TryGetValue("bridgetokubernetes/ignore-ports", out var ports)) {
+                        var allPorts = ports.Split(",");
+                        foreach(var port in allPorts) {
+                            try {
+                                portsToIgnore.Add(int.Parse(port));
+                            } catch {
+                                throw new Exception("The ports are not in the correct format");
```
elenavillamil, line 134: Let's add in this message the ports string user used as well as well as some guidance on how to fix. Something like:
("bridgetokubernetes/ignore-ports configuration input with value {1} is invalid. It should be a comma separated list of integer ports")

I also think it should be an user visible exception: UserVisibleException, it may require you to add the string format to Resources. You can use line 113 as an example.

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -117,6 +120,28 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
...
+                            }
+                            
+                        }
+                    }
+                }
+            }
+            _log.Info("Ignoring ports: {0}", portsToIgnore.Count > 0 ? portsToIgnore.ToString() : "none");
```
elenavillamil, line 141: Maybe something more explicit on this log. For example: 

"Per bridgetokubernetes/ignore-ports user config the following ports will be ignored: {0}"

elenavillamil (commented):
Change looks good, left some minor comments. Also, we already have UT coverage for GetReachableEndpointsAsync, can you update the UTs or add new UTs following existing one as example to cover this new code path?

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -117,6 +121,27 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
             }
         }
 
+        //
+        public List<int> GetPortsToIgnoreFromAnnotations(V1ServiceList serviceNameSpace) {
```
hsubramanianaks, line 125: @Vidya2606 can you add a unit test case for this method, since it is public method, it should be easily testable. Thank you for the PR.

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -117,6 +121,27 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
             }
         }
 
+        //
+        public List<int> GetPortsToIgnoreFromAnnotations(V1ServiceList serviceNameSpace) {
+            var portsToIgnore = new List<int>();
+            foreach(var item in serviceNameSpace.Items) {
+                if (item.Metadata != null && item.Metadata.Annotations != null) {
+                    if (item.Metadata.Annotations.TryGetValue("bridgetokubernetes/ignore-ports", out var ports)) {
```
Tatsinnit, line 129: 💡 since we are using `split` I am guessing it guarantee `string type` so lets explicitly mention that, `out string ports`?

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -117,6 +121,27 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
             }
         }
 
+        //
+        public List<int> GetPortsToIgnoreFromAnnotations(V1ServiceList serviceNameSpace) {
+            var portsToIgnore = new List<int>();
+            foreach(var item in serviceNameSpace.Items) {
```
Tatsinnit, line 127: 💡 Thoughts: Can we make this `for each` simpler for maintainability of when this funct will get added with more in future.

There is nested `foreach loop` in this can we `single responsibility` so one inner iteration `IgnorePorts()` loops through and make it replaceable.

Tatsinnit (commented):
☕️❤️💡🙏 Thank you for adding me into this review, here are some initial thoughts and ideas. thank you.

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -117,6 +121,27 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
...
+                            try {
+                                portsToIgnore.Add(int.Parse(port));
+                            } catch {
+                                throw new UserVisibleException(this._operationContext, $"bridgetokubernetes/ignore-ports configuration value {port} is invalid. It should be a comma separated list of integer ports");
+                            }
+                        }
+                    }
+                }
+            }
+            _log.Info("Per bridgetokubernetes/ignore-ports user config the following ports will be ignored: {0}", portsToIgnore.Count > 0 ? portsToIgnore.ToString() : "none");
```
hsubramanianaks, line 141: hi Vidya, just a suggestion to improvise this, you can use where (which filters items based on some conditions) and avoid extra for each (nested) by using string explicitly and toList(). Thanks. I didn't test this, please make sure to test this if you are planning to use them :).
```suggestion
            _public List<int> GetPortsToIgnoreFromAnnotations(V1ServiceList serviceNameSpace) {
            var portsToIgnore = new List<int>();
            // only filter the services which has this annotation present
            List<V1Service> filteredList = serviceNameSpace.Items.Where(item => item.Metadata != null && item.Metadata.Annotations != null 
                && item.Metadata.Annotations.ContainsKey("bridgetokubernetes/ignore-ports")).ToList();

            // loop through them
            filteredList.ForEach(item =>
            {
                if (item.Metadata.Annotations.TryGetValue("bridgetokubernetes/ignore-ports", out string ports))
                {
                    if (ports.Length > 0)
                    {
                        try {
                            // this avoids extra for each loop
                            portsToIgnore = ports.Split(",").ToList();
                        }
                        catch
                        {
                            throw new UserVisibleException(this._operationContext, $"bridgetokubernetes/ignore-ports configuration value {port} is invalid. It should be a comma separated list of integer ports");
                        }
                    }
                    
                }

            });
            _log.Info("Per bridgetokubernetes/ignore-ports user config the following ports will be ignored: {0}", portsToIgnore.Count > 0 ? portsToIgnore.ToString() : "none");
            return portsToIgnore;
        }

```

src/library/Connect/WorkloadInformationProvider.cs
Vidya2606: I have simplified the func which avoids the use of split.

src/library.tests/WorkloadInformationProviderTests.cs
```diff
@@ -47,63 +47,97 @@ public async void GetReachableServicesAsync_HeadlessService(int numServices, int
...
-            bool addSubset = false;
-            for (int i = 0; i < numServices; i++)
+        [Theory]
+        [InlineData(1, 100)]
+        [InlineData(5, 20)]
+        [InlineData(10, 3)]
+        [InlineData(1, 3)]
+        public async void GetReachableServicesAsync_PortsToIgnore_HeadlessService(int numServices, int numAddresses)
+        { 
+            List<int> portsToIgnore = new List<int> { 190, 570 };
```
elenavillamil, line 57: I believe this test right now is only testing the scenario where none of hte user ports match what the user configured in bridgetokubernetes/ignore-ports. For test to exercise all scenario you need to add these two ports or at least one of them to the input set (aka subsets list in that function). When ConfigureHeadlesService function runs, we need to make sure that on the response we are mocking for GetEndpointInNamespaceAsync there is some endpoint with the port that the user would want to ignore.

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -117,6 +122,31 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
             }
         }
 
+        public List<int> GetPortsToIgnoreFromAnnotations(V1ServiceList serviceNameSpace) {
+            var portsToIgnore = new List<int>();
+
+            List<V1Service> portsList = serviceNameSpace.Items.Where(item => item.Metadata != null && 
+            item.Metadata.Annotations != null &&
+            item.Metadata.Annotations.ContainsKey(ServiceAnnotations)).ToList();
```
elenavillamil, line 130: nit: I believe current code style uses ? for null checks. So something like where(item => item.Metadata?.Annotations?.ContainsKey). More details on this operator [here](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-)

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -117,6 +122,31 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
...
 
+        public List<int> GetPortsToIgnoreFromAnnotations(V1ServiceList serviceNameSpace) {
+            var portsToIgnore = new List<int>();
+
+            List<V1Service> portsList = serviceNameSpace.Items.Where(item => item.Metadata != null && 
+            item.Metadata.Annotations != null &&
+            item.Metadata.Annotations.ContainsKey(ServiceAnnotations)).ToList();
+
+            portsList.ForEach(port => {
+                if (port.Metadata.Annotations.TryGetValue(ServiceAnnotations, out string ports)) {
```
elenavillamil, line 133: I think we should probably add null check on Metadata and Annotations here:
if (port.Metadata?.Annotations?.TryGetValue)

src/library.tests/WorkloadInformationProviderTests.cs
```diff
@@ -47,63 +47,97 @@ public async void GetReachableServicesAsync_HeadlessService(int numServices, int
...
+                    Metadata = new V1ObjectMeta()
+                    {
+                        Name = namingFunction(i),
+                    }
+                };
+
+                if (!string.IsNullOrEmpty(ignorePorts))
+                {
+                    service.Metadata.Annotations = new Dictionary<string, string>();
+                    service.Metadata.Annotations.Add("bridgetokubernetes/ignore-ports", ignorePorts);
```
elenavillamil, line 98: Aside from when ignore-ports is present with valid value, lets add coverage for when ignore-ports is empty, not present at all, or present with bad formatted string

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -117,6 +122,31 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
...
+                    if (ports.Length > 0) {
+                        try {
+                            // ports e.g. "460, 543"
+                            var ignorePorts = ports.Split(",").Select(p => int.Parse(p));
+                            portsToIgnore.AddRange(ignorePorts);
+                        } 
+                        catch {
+                            throw new UserVisibleException(this._operationContext, $"bridgetokubernetes/ignore-ports configuration value {port} is invalid. It should be a comma separated list of integer ports");
```
elenavillamil, line 141: typo, i think this needd to be ports. Once you add UT for bad formatted string it should catch it :)

elenavillamil: I believe once you add UT, you will see that usage of UserVisableException is wrong, it required some pre-defined format. I think it may need to follow format similar to line 102 ("throw new UserVisibleException(_operationContext, Resources.PortRequiredFormat, externalEndpoint.Name);")

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -117,6 +122,27 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
...
 
+        public List<int> GetPortsToIgnoreFromAnnotations(V1ServiceList serviceNameSpace) {
+            var portsToIgnore = new List<int>();
+            List<V1Service> portsList = serviceNameSpace.Items.Where(item => item.Metadata?.Annotations?.ContainsKey(ServiceAnnotations) ?? false).ToList();
+            portsList.ForEach(port => {
+                if (port.Metadata?.Annotations?.TryGetValue(ServiceAnnotations, out string ports) ?? false) {
+                    if (ports.Length > 0) {
+                        try {
+                            // ports e.g. "460, 543"
+                            var ignorePorts = ports.Split(",").Select(p => int.Parse(p));
```
hsubramanianaks, line 133: Hi Vidya we can still make this one liner avoiding a local variable, if it makes things complex, please ignore it. 

```suggestion
                            portsToIgnore =ports.Splt(",")?.Select(Int32::Parse)?.toList()

```

sabbour (changes requested):
@Vidya2606 can you please add a snippet to the pull request and possibly [README.md](https://github.com/Azure/Bridge-To-Kubernetes/blob/main/README.md) showing what annotations to use on the service to trigger this functionality? This will help with updating the documentation.

sabbour (approved):
Looks good to me from a functional standpoint.

src/library.tests/WorkloadInformationProviderTests.cs
```diff
@@ -1,109 +1,174 @@
...
+                }
+            }
+            Assert.True(portsInResult.All(p => portsOffered.Contains(p)));
+        }
+
+        [Theory]
+        [InlineData(5, 20)]
+        public async void GetReachableServicesAsync_PortsToIgnoreIncorrectFormat_HeadlessService(int numServices, int numAddresses)
+        { 
+            ConfigureHeadlessService(numServices: numServices, namingFunction: (i) => $"myapp-{i}", numAddresses: numAddresses, addressHostNamingFunction: (i) => $"Host-{i}", "ajhfja, 19o");
```
elenavillamil, line 100: nit: keep same format as rest of function call:
ignorePorts: "ajhfja, 19o");

elenavillamil: same in other places that make this function call.

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -34,6 +34,7 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
         private readonly IResourceNamingService _resourceNamingService;
         private AsyncLazy<RemoteContainerConnectionDetails> _remoteContainerConnectionDetails;
         private readonly IProgress<ProgressUpdate> _progress;
+        private const string ServiceAnnotations = "bridgetokubernetes/ignore-ports";
```
elenavillamil, line 37: nit: check if constants file may be better place for this const. For example similar to: ingress?.Metadata?.Annotations?.ContainsKey(Constants.Https.CertManagerAnnotationKey)

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -117,6 +122,26 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
             }
         }
 
+        public List<int> GetPortsToIgnoreFromAnnotations(V1ServiceList serviceNameSpace) {
+            var portsToIgnore = new List<int>();
+            List<V1Service> portsList = serviceNameSpace.Items.Where(item => item.Metadata?.Annotations?.ContainsKey(ServiceAnnotations) ?? false).ToList();
```
elenavillamil, line 127: Doesn't ContainsKey already return a boolen? why the "?? false"?

elenavillamil, line 127: this name seems wrong, should it be servicesList or servicesWithIgnorePorts?

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -117,6 +122,26 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
             }
         }
 
+        public List<int> GetPortsToIgnoreFromAnnotations(V1ServiceList serviceNameSpace) {
+            var portsToIgnore = new List<int>();
+            List<V1Service> portsList = serviceNameSpace.Items.Where(item => item.Metadata?.Annotations?.ContainsKey(ServiceAnnotations) ?? false).ToList();
+            portsList.ForEach(port => {
```
elenavillamil, line 128: once you rename the variable this should be "forEach(service => {" instead of "ForEach(port =>"

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -117,6 +122,26 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
...
 
+        public List<int> GetPortsToIgnoreFromAnnotations(V1ServiceList serviceNameSpace) {
+            var portsToIgnore = new List<int>();
+            List<V1Service> portsList = serviceNameSpace.Items.Where(item => item.Metadata?.Annotations?.ContainsKey(ServiceAnnotations) ?? false).ToList();
+            portsList.ForEach(port => {
+                if (port.Metadata?.Annotations?.TryGetValue(ServiceAnnotations, out string ports) ?? false) {
+                    if (ports.Length > 0) {
+                        try {
+                            // ports e.g. "460, 543"
+                            portsToIgnore = ports.Split(",").Select(Int32.Parse)?.ToList();
```
elenavillamil, line 133: I dont think this works if there is more than one service in the namespace. It will get overriden instead of append. Also I think we need to keep track of what ports for each service. for example if there are 3 services in the namespace and the config of service A says ignore port 80, serivce B says ignore port 300, service C says ignore port 3000. Then serivce A configures port 80 and 300, for service A we should only ignore port 80 not both.

src/library.tests/WorkloadInformationProviderTests.cs
```diff
@@ -1,109 +1,174 @@
...
+            foreach (var endpoint in result)
+            {
+                if (endpoint.Ports.Any())
+                {
+                    // None of the ports to ignore must be used 
+                    Assert.Empty(endpoint.Ports.Where(p => portsToIgnore.Contains(p.RemotePort)));                    
+                    portsInResult.AddRange(endpoint.Ports.Select( p => p.RemotePort));
+                }
+            }
+            Assert.True(portsAllocated.All(p => portsInResult.Contains(p)));
```
elenavillamil, line 74: how does this check that portsInResult does not contain ignore ports? I think this line is making sure that all ports inside PortsAllocated are inside portsInResult, but if portsInresult has extra entries we wont catch it, will we?

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -382,7 +407,8 @@ private int GetServicePortValue(V1ServicePort servicePort, V1Container container
         private async Task<IEnumerable<EndpointInfo>> _CollectServicesToRouteAsync(
             string workloadNamespace,
             IEnumerable<V1Service> services,
-            CancellationToken cancellationToken)
+            CancellationToken cancellationToken,
+            List<int> portsToIgnore)
```
elenavillamil, line 411: I think this has to be a map of service to ports to ignore for the service. Else if user adds port 80 to be ignore for service A, we will ignore port 80 for all services and that does not seem correct behavior.

src/library.tests/WorkloadInformationProviderTests.cs
```diff
@@ -1,109 +1,174 @@
...
+                    },
+                    Metadata = new V1ObjectMeta()
+                    {
+                        Name = namingFunction(i),
+                    }
+                };
+
+                if (!string.IsNullOrEmpty(ignorePorts))
+                {
+                    service.Metadata.Annotations = new Dictionary<string, string>();
```
elenavillamil, line 128: this add all same ports to ignore for every service. My guess is that user would configure different ports to ignore for different services.

src/library/Connect/WorkloadInformationProvider.cs
Vidya2606: Cannot implicitly convert type 'bool?' to 'bool'. 
bool? takes in true, false and null. So explicit conversion is required.

src/library.tests/WorkloadInformationProviderTests.cs
Vidya2606: I'm checking this in line 70, which is making sure that non of the portsToIgnore appear in the results. 
I will be re-writing this to test the updated logic.

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -431,10 +462,11 @@ private int GetServicePortValue(V1ServicePort servicePort, V1Container container
 
             var servicesToRouteEndpointInfos = servicesToRouteMap.Values.Select(s => new EndpointInfo()
             {
+                // portsToIgnore is a directory which contains a separate list of ports to ignore for each service, when picking ports for a service any port which is in the portsToIgnore is not used.
```
elenavillamil, line 465: nit: move this comment to line below?

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -458,8 +490,8 @@ private int GetServicePortValue(V1ServicePort servicePort, V1Container container
                         {
                             DnsName = isInWorkloadNamespace ?
                                         $"{address.Hostname}.{endpoint.Metadata.Name}" :
-                                        $"{address.Hostname}.{endpoint.Metadata.Name}.{endpoint.Metadata.Namespace()}",
-                            Ports = subset.Ports?.Where(port => this._IsSupportedProtocol(port.Protocol, endpoint.Metadata.Name)).Select(p => new PortPair(remotePort: p.Port)).ToArray() ?? new PortPair[] { },
+                                        $"{address.Hostname}.{endpoint.Metadata.Name}.{endpoint.Metadata.Namespace()}", // check the below code
```
elenavillamil, line 493: nit: remove that comment at the end?

src/library/Constants.cs
```diff
@@ -10,6 +10,7 @@ internal static class Constants
         internal static class AzureDevSpacesService
         {
             public const string UserClusterNamespaceName = "azds";
+            public const string ServiceAnnotations = "bridgetokubernetes/ignore-ports";
```
elenavillamil, line 13: nit: i would make its own class of DeploymentConfig, something like:
internal static class DeploymentConfig
{
       public const string ServiceAnnotations = "bridgetokubernetes/ignore-ports";
}

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -117,6 +122,25 @@ internal class WorkloadInformationProvider : IWorkloadInformationProvider
             }
         }
 
+        public Dictionary<String, IList<int>> GetPortsToIgnoreFromAnnotations(V1ServiceList serviceNameSpace) {
+            Dictionary<String, IList<int>> servicePortsToIgnore = new Dictionary<String, IList<int>>();
+            List<V1Service> servicesWithIgnorePorts = serviceNameSpace.Items.Where(item => item.Metadata?.Annotations?.ContainsKey(DeploymentConfig.ServiceAnnotations) ?? false).ToList();
+            servicesWithIgnorePorts.ForEach(service => {
+                if (service.Metadata?.Annotations?.TryGetValue(DeploymentConfig.ServiceAnnotations, out string ports) ?? false) {
+                    if (ports.Length > 0) {
```
Tatsinnit, line 130: 💡 We have the value of `ports` above, why are we waiting until now - we could do better error handling by failing early scenarios like:

```.cs
if (ports.Length === 0) {  // error handle stuff here }
```

Vidya2606: The try catch is for handling any parsing errors.
ports.length == 0 is a valid condition.

elenavillamil:
> @Vidya2606 can you please add a snippet to the pull request and possibly [README.md](https://github.com/Azure/Bridge-To-Kubernetes/blob/main/README.md) showing what annotations to use on the service to trigger this functionality? This will help with updating the documentation.

@sabbour, will you be able to update public docs as well once this is rolled out?

src/library.tests/WorkloadInformationProviderTests.cs
```diff
@@ -1,109 +1,261 @@
...
+        private static bool CompareLists<T>(List<T> list1, List<T> list2) 
+        {
+            if (list1.Count != list2.Count)
+            {
+                return false;
+            }
+            return list1.All( item => list2.Contains(item));
+        }
+
+        private void ConfigureHeadlessServiceWithIgnorePorts(int numAddresses, Dictionary<string, string> ignorePorts, Dictionary<string, List<int>> servicePorts = null)
```
elenavillamil, line 145: I think we should rename this function to ConfigureServiceWithIgnorePorts and ensure we add both types of services otherwise as it is now only half of the code your added is being tested (aka only the headlessservice code path is currently executed).

elenavillamil: you will want to also rename the UTs to not mention they are only for headless service

elenavillamil (approved):
Looks good. UT coverage is great now, thanks!

hsubramanianaks (approved):
LGTM. Thank you.

PR #149, opened 2023-01-20
Update azure-pipelines.yml for Azure Pipelines

elenavillamil:
What this PR does: If PR triggered pipeline (sign type test), then we will exclude all the signing tasks. 

This is first step to splitting pipelines into two. One build pipeline for PRs and one build pipeline for release (this will be a clone of build pipelines + signing). This second pipeline will remain internal (not github exposed) and this way we can ensure that sign build will only run on main and started by Bridge team members.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -97,12 +97,12 @@ jobs:
     name: ''
     displayName: Install Signing Plugin
     enabled: True
-    condition: eq(parameters.SignTypeToUse, 'real')
+    condition: eq(${{ parameters.SignTypeToUse }}, 'real')
     inputs:
       signType: ${{ parameters.SignTypeToUse }}
       zipSources: false
   - task: MicroBuildLocalizationPlugin@3
-    condition: eq(parameters.SignTypeToUse, 'real')
+    condition: eq(${{ parameters.SignTypeToUse }}, 'real')
     displayName: 'Install Localization Plugin'
     enabled: True
   - task: UseDotNet@2
@@ -272,7 +272,7 @@ jobs:
     name: ''
     displayName: Sign Files
     enabled: True
-    condition: eq(parameters.SignTypeToUse, 'real')
+    condition: eq(${{ parameters.SignTypeToUse }}, 'real')
     inputs:
       solution: src/client.signproj
       msbuildArgs: /t:SignFiles /p:SignType=${{ parameters.SignTypeToUse }}
@@ -563,12 +563,12 @@ jobs:
   - task: MicroBuildSigningPlugin@3
     displayName: Install Signing Plugin
     enabled: True
-    condition: eq(parameters.SignTypeToUse, 'real')
+    condition: eq(${{ parameters.SignTypeToUse }}, 'real')
     inputs:
       signType: ${{ parameters.SignTypeToUse }}
       zipSources: false
   - task: MicroBuildLocalizationPlugin@3
-    condition: eq(parameters.SignTypeToUse, 'real')
+    condition: eq(${{ parameters.SignTypeToUse }}, 'real')
     displayName: Install Localization Plugin
   - task: UseDotNet@2
     displayName: Use .NET Core SDK 6.0.X
@@ -672,7 +672,7 @@ jobs:
   - task: VSBuild@1
     displayName: Sign Files
     enabled: True
-    condition: eq(parameters.SignTypeToUse, 'real')
+    condition: eq(${{ parameters.SignTypeToUse }}, 'real')
     inputs:
       solution: src/client.signproj
       msbuildArgs: /t:SignFiles /p:SignType=${{ parameters.SignTypeToUse }}
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -78,9 +78,6 @@ jobs:
             Write-Host "Done setting the sign type and telemetry key instance for main branch"
         }
 
-
-        Write-Host "Disabling NuGet package signing because it doesn't work, and we don't publish public NuGet packages"
-
         Write-Host "##vso[task.setvariable variable=MicroBuild_NuPkgSigningEnabled]true"
   - task: MicroBuildSigningPlugin@3
     name: ''
@@ -542,8 +539,6 @@ jobs:
             Write-Host "##vso[task.setvariable variable=TelemetryType]"{{ parameters.TelemetryType }}
             Write-Host "Done setting the sign type and telemetry key instance for staging branch"
         }
-
-        Write-Host "Disabling NuGet package signing because it doesn't work, and we don't publish public NuGet packages"
 
         Write-Host "##vso[task.setvariable variable=MicroBuild_NuPkgSigningEnabled]true"
   - task: MicroBuildSigningPlugin@3
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -97,12 +97,12 @@ jobs:
     name: ''
     displayName: Install Signing Plugin
     enabled: True
-    condition: eq(${{ parameters.SignTypeToUse }}, 'real')
+    condition: eq('${{ parameters.SignTypeToUse }}', 'real')
     inputs:
       signType: ${{ parameters.SignTypeToUse }}
       zipSources: false
   - task: MicroBuildLocalizationPlugin@3
-    condition: eq(${{ parameters.SignTypeToUse }}, 'real')
+    condition: eq('${{ parameters.SignTypeToUse }}', 'real')
     displayName: 'Install Localization Plugin'
     enabled: True
   - task: UseDotNet@2
@@ -272,7 +272,7 @@ jobs:
     name: ''
     displayName: Sign Files
     enabled: True
-    condition: eq(${{ parameters.SignTypeToUse }}, 'real')
+    condition: eq('${{ parameters.SignTypeToUse }}', 'real')
     inputs:
       solution: src/client.signproj
       msbuildArgs: /t:SignFiles /p:SignType=${{ parameters.SignTypeToUse }}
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -97,10 +97,12 @@ jobs:
     name: ''
     displayName: Install Signing Plugin
     enabled: True
+    condition: eq(parameters.SignTypeToUse, 'real')
```
danegsta, line 100: '${{ parameters.SignTypeToUse }}'

elenavillamil: thanks :)

azure-pipelines.yml
```diff
@@ -97,10 +97,12 @@ jobs:
     name: ''
     displayName: Install Signing Plugin
     enabled: True
+    condition: eq('${{ parameters.SignTypeToUse }}', 'real')
```
Tatsinnit, line 100: ☕️💡 thought share: Since this signin part was alsonpartnofnone bug against the repo, I wonder if we don’t differentiate between test and real for sign in so that test can be as close to real and tested.

what do you think @hsubramanianaks and @elenavillamil ! ☕️☕️🕵️‍♀️

Tatsinnit (commented):
☕️💡 please note: (just 5 cents worth of thought) I think the cli for this PR is broken because of the new checks introduced probably producing non-singed files. Re: test- I have mentioned in other comment that we should keep production and test very close beside test name different unless we have a valid reasons not to sign the binaries for test environments:


6BD2D159-D968-440F-8E28-940B5472BFD2

azure-pipelines.yml
```diff
@@ -97,10 +97,12 @@ jobs:
...
     inputs:
       signType: ${{ parameters.SignTypeToUse }}
       zipSources: false
   - task: MicroBuildLocalizationPlugin@3
+    condition: eq('${{ parameters.SignTypeToUse }}', 'real')
```
hsubramanianaks, line 105: Hi, just an idea we can keep the pipelines file single and reuse them for PR validations and PR with signed files. I think splitting them and creating duplicates can be avoided. You can refer VS extension pipelines for reference which does similar approach. Also, another point which Tats also pointed out is "test" sign the binaries for PR's instead of real signing. Is it worth to check with other teams what's the use of "test" sign vs "real" sign? so that we don't need to skip test signing.

we can also figure out the auth issue for installing signing plugin. My thought on the auth issue for installing signing plugin is to give access to GitHub service connection/account so that it is authorized to install the plugin at least for the test signs (we can give a try on this), for real signing agreed we need to go through the M2 approval process.

hsubramanianaks:
> ☕️💡 please note: (just 5 cents worth of thought) I think the cli for this PR is broken because of the new checks introduced probably producing non-singed files. Re: test- I have mentioned in other comment that we should keep production and test very close beside test name different unless we have a valid reason not to sign the binaries for test environments:
> 
> 6BD2D159-D968-440F-8E28-940B5472BFD2

Yes, B2K Nuget pipeline will fail since we are skipping signing and it verifies for signing. There needs to be changes to [this pipeline](https://devdiv.visualstudio.com/DevDiv/_release?_a=releases&view=mine&definitionId=3330) to make successful and someone has to trigger all these pipelines manually instead of automatically triggering them. Please update release wiki and also can we discuss about this if we are going down this approach further? because now if we are changing this model it needs to be triggered manually every time someone wants their changes in "DEV" or "Staging" environment. **Note:** B2K Images pipeline can be made manual since that was causing issues in the past, but my thought is it would cause issues only if there are potential breaking changes in one of the components (devhost, restoration and routing manager) that causes CLI to pick wrong images, which it doesn't happen often. Please let me know your thoughts. Thank you.

azure-pipelines.yml
elenavillamil: Guidance from David was that PR builds should not have signing (for security reasons to avoid signing something sent from someone outside our team since this is a public repo - specially after this pipeline is approved for real signing). Plan is to have one build pipeline for PRs and one "official" build pipeline with signing for release. This is step one of refactoring. Once this works I am working on pipeline that fully removes signing that will be used with PRs. This pipeline will be unlinked from PRs and always have true signing (potentially run it of main once a day). PR description has details as well. Let me know if this doesnt address the concern.

azure-pipelines.yml
elenavillamil: Thanks Hari. I synced on this with David from DevDiv last week before sending this PR. It was his advice to separate pipelines (reply to one of tats comments on the why as well as PR description has more details). 

"test" is a mode to make sure everything is ready for signing but it wont use real valid certs and actually sign. The usage is to run as close to signing as possible but saving time since real signing is not done. I think this is for cases where signing takes a long time (normally big binaries to sign - singing step of the pipeline in one of my previous teams took 40+ min). This is not very useful to us since we have small binaries and signing step does not add that much time for us. So even if we move signing out to separate PR, I do not think we will use test option.

azure-pipelines.yml
danegsta: As a note on the linked issue, it looks like the problem was that a test signed build actually passed the signing validation check and was accidentally shipped with a VSCode extension release. Test signing in these pipelines doesn't actually sign with a valid trusted certificate and sign check validation should ideally fail for a test signed build as well as an unsigned build.

elenavillamil:
> > ☕️💡 please note: (just 5 cents worth of thought) I think the cli for this PR is broken because of the new checks introduced probably producing non-singed files. Re: test- I have mentioned in other comment that we should keep production and test very close beside test name different unless we have a valid reason not to sign the binaries for test environments:
> > 6BD2D159-D968-440F-8E28-940B5472BFD2
> 
> Yes, B2K Nuget pipeline will fail since we are skipping signing and it verifies for signing. There needs to be changes to [this pipeline](https://devdiv.visualstudio.com/DevDiv/_release?_a=releases&view=mine&definitionId=3330) to make successful and someone has to trigger all these pipelines manually instead of automatically triggering them. Please update release wiki and also can we discuss about this if we are going down this approach further? because now if we are changing this model it needs to be triggered manually every time someone wants their changes in "DEV" or "Staging" environment. **Note:** B2K Images pipeline can be made manual since that was causing issues in the past, but my thought is it would cause issues only if there are potential breaking changes in one of the components (devhost, restoration and routing manager) that causes CLI to pick wrong images, which it doesn't happen often. Please let me know your thoughts. Thank you.

Hi Hari and Tats. The concern makes sense, but there are two things worth nothing that I think address the concern:
1. This was already broken. This is main reason behind this change - every single PR now that is sent form a fork will always have 3 failed checks. I felt this is a quality issue since having this checks is not adding value if they are always failing. When I looked at option of keeping singing, I synced with David on his thoughts. this is when he explained concerned I mentioned above of having pipeline with signing on a PRs for a public repo. Hence this direction.
2. Before we had to manually go an trigger the build to test on dev anyway (reason for manual trigger is because the PR trigger fails), as Hari mentioned from that point on everything (docker images and nuget) would be automated. After this change we would still need one manual trigger of the build pipeline but my idea is it would be the release build pipeline (aka with signing), and from there everything will be automated in to dev like it is now. Thus end result should be same as now in terms of manual triggers but PRs will have 2 less failed checks and we will declouple signing from PR validation. Thus improving quality and security.

hsubramanianaks:
> > > ☕️💡 please note: (just 5 cents worth of thought) I think the cli for this PR is broken because of the new checks introduced probably producing non-singed files. Re: test- I have mentioned in other comment that we should keep production and test very close beside test name different unless we have a valid reason not to sign the binaries for test environments:
> > > 6BD2D159-D968-440F-8E28-940B5472BFD2
> > 
> > 
> > Yes, B2K Nuget pipeline will fail since we are skipping signing and it verifies for signing. There needs to be changes to [this pipeline](https://devdiv.visualstudio.com/DevDiv/_release?_a=releases&view=mine&definitionId=3330) to make successful and someone has to trigger all these pipelines manually instead of automatically triggering them. Please update release wiki and also can we discuss about this if we are going down this approach further? because now if we are changing this model it needs to be triggered manually every time someone wants their changes in "DEV" or "Staging" environment. **Note:** B2K Images pipeline can be made manual since that was causing issues in the past, but my thought is it would cause issues only if there are potential breaking changes in one of the components (devhost, restoration and routing manager) that causes CLI to pick wrong images, which it doesn't happen often. Please let me know your thoughts. Thank you.
> 
> Hi Hari and Tats. The concern makes sense, but there are two things worth nothing that I think address the concern:
> 
> 1. This was already broken. This is main reason behind this change - every single PR now that is sent form a fork will always have 3 failed checks. I felt this is a quality issue since having this checks is not adding value if they are always failing. When I looked at option of keeping singing, I synced with David on his thoughts. this is when he explained concerned I mentioned above of having pipeline with signing on a PRs for a public repo. Hence this direction.
> 2. Before we had to manually go an trigger the build to test on dev anyway (reason for manual trigger is because the PR trigger fails), as Hari mentioned from that point on everything (docker images and nuget) would be automated. After this change we would still need one manual trigger of the build pipeline but my idea is it would be the release build pipeline (aka with signing), and from there everything will be automated in to dev like it is now. Thus end result should be same as now in terms of manual triggers but PRs will have 2 less failed checks and we will declouple signing from PR validation. Thus improving quality and security.

is this the plan moving forward , can you please confirm?

	Trigger PR pipelines without signing (because test signing is not worth it, seems like from David's comment) when a PR is raised against main branch and deploy/release to Dev branch. 
	Trigger Main branch release pipelines with signing (real) when a PR is merged to main branch and deploy/release to staging.
	B2K Images will be manually triggered for each PR and deployed to Dev environment? is that correct? my concern is how are we controlling images CI/CD process. I would suggest manual for now and see how it goes from there. It needs to run only when there are changes in the images. 
	Binaries signing verification fails today if binaries are NOT signed, but we need to update them to fail if it is NOT real signed. 

Also maintain single file for common steps in the pipelines instead of duplication and reuse them as much as possible. 

Please correct me If I have missed anything,Thanks Elena for driving this.

azure-pipelines.yml
elenavillamil: resolving based on offline sync, feel free to re-open if needed.

azure-pipelines.yml
elenavillamil: resolving based on offline sync. Please feel free to re-open if needed.

hsubramanianaks (approved):
LGTM, if all builds passes, we can merge this.

PR #154, opened 2023-01-26
Set up build pipeline for PRs

elenavillamil:
Creating build pipeline for PRs that does not have signing.


hsubramanianaks (approved):
LGTM, I compared with existing pipelines, this is just minus the signing plugin and sign tasks for both self contained and non self contained jobs. Thank you.

hsubramanianaks (approved):
LGTM, reapproving again.

PR #163, opened 2023-02-11
Resolves a issue regaring watching of deployments

### LINKED ISSUE #162: Automatic restart of RemoteAgent does not work because of missing parameter in code
The automatic restart of the agent does not work because the functionallity that watches the deployment for changes is missing a parameter.

This row:
https://github.com/Azure/Bridge-To-Kubernetes/blob/59c7d82d03668b322d9270f497e31ac7b6521333/src/common/Kubernetes/KubernetesClient.cs#L192

Is missing the `watch: true` parameter and that causes the following warning to occure in the logfile:
`Library | WARNG | Watching 'default/MyService' failed, error not a watchable request or failed response.`

This in turn stops the watching of deployments so automatic restart of RemoteAgent will never happen.

The row can be changed to this to solve this issue:
`return ClientInvokeWrapperAsync(async () => await RestClient.AppsV1.ListNamespacedDeploymentWithHttpMessagesAsync(namespaceName, fieldSelector: $"metadata.name={deploymentName}", timeoutSeconds: timeoutSeconds, watch: true, cancellationToken: cancellationToken), nameof(WatchV1DeploymentAsync), cancellationToken);`

I have submited a PR that fixes this.

Eneuman:
This PR fixes a bug that causes watching of deployments not to work.

There is a comment here:
https://github.com/Azure/Bridge-To-Kubernetes/blob/e0fd5849780d89d948f7d4e90eef06eaa5238246/src/library/Connect/KubernetesRemoteEnvironmentManager.cs#L237

Maby this can be removed now :)

closes #162


elenavillamil:
Thanks for submitting this @Eneuman! Sending it to our testing team, will approve once we have green light. I am assuming now warning is gone from log file right?

Eneuman:
@elenavillamil Yeah, warning is gone and it seems to start faster aswell :)

elenavillamil (approved):
Change looks good and we also got green light from testing team. Thank you @Eneuman !

hsubramanianaks (approved):
LGTM.

PR #164, opened 2023-02-11
New feature: Disable probes in remote agent

### LINKED ISSUE #111: Liveliness/readiness probe - bridge container being recycled
**Describe the bug**
Container with liveliness (readiness) probe - when connected to, probe is not removed from the spec, container will be restarted in a loop

**Mention the platform you are using**
VS 2022 Enterprise - 17.4

**To Reproduce**

Service spec has a liveliness probe e.g.

```yaml
livenessProbe:
      httpGet:
        path: /health
        port: 3000
      initialDelaySeconds: 5
      periodSeconds: 5
```

Connect bridge in VS.. this overwrites service spec to point to bridge container image
probe is not "disabled"
K8 attempts check the probe, this fails against the bridge image and container is being restarted

**Expected behavior**

liveliness/readiness probe "parked" and restored when the bridge is disconnected

or 

document the limitation - I can't see currently any limitation documented re this

**Desktop (please complete the following information):**
 - OS: W10

### LINKED ISSUE #114: Remote agent container fails readiness probes
At first I’ve posted this issue to vscode extension repo, but probably it’s the right repo for this issue, so I’m just copying it here. Sorry in advance for opening it in both repos. 

**Describe the bug**
I'm trying to use Bridge to Kubernetes as described in [this document](https://learn.microsoft.com/en-us/visualstudio/bridge/bridge-to-kubernetes-vs-code#install-and-use-local-tunnel-debugging). I was able to start tunneling of my service. However, the new pod that is created with remote agent container, never becomes ready, since it fails readiness probe.

If I run `kubectl describe` on a pod that Bridge tries to create, I see that it fails readiness probe:

```
Events:                                                                                                                                                                    │
│   Type     Reason     Age                  From               Message                                                                                                      │
│   ----     ------     ----                 ----               -------                                                                                                      │
│   Normal   Scheduled  12m                  default-scheduler  Successfully assigned p-buchyn/dev-users-frontend-main-696897f9ff-f6w5g to ip-10-0-92-131.ec2.internal       │
│   Normal   Pulling    12m                  kubelet            Pulling image "bridgetokubernetes.azurecr.io/lpkremoteagent:0.1.8"                                           │
│   Normal   Pulled     12m                  kubelet            Successfully pulled image "bridgetokubernetes.azurecr.io/lpkremoteagent:0.1.8" in 2.012129165s               │
│   Normal   Created    12m                  kubelet            Created container app                                                                                        │
│   Normal   Started    12m                  kubelet            Started container app                                                                                        │
│   Warning  Unhealthy  2m5s (x66 over 11m)  kubelet            Readiness probe failed: Get "http://10.0.94.175:80/": context deadline exceeded (Client.Timeout exceeded whi │
│ le awaiting headers)
```

`10.0.94.175` here is the IP of this pod itself.

The logs of remote agent container look like follows (I post only a part of the logs, but this logs seem to be repeated over time):

```
2022-11-22T23:31:34.2939966Z | RemoteAgent | TRACE | PortForwardConnector.Disconnect 168                                                                                   │
│ 2022-11-22T23:31:34.2940595Z | RemoteAgent | TRACE | AgentHub closed for 80, id 168                                                                                        │
│ 2022-11-22T23:31:38.8318380Z | RemoteAgent | TRACE | ReversePortForwardConnector on port 80 accepted incoming request as stream 169.                                       │
│ 2022-11-22T23:31:38.8320319Z | RemoteAgent | TRACE | AgentHub connnected for 80, id 169                                                                                    │
│ 2022-11-22T23:31:38.8321460Z | RemoteAgent | TRACE | ReversePortForwardConnector.HookupStreamData ReadAsync 169 returns 121 bytes.                                         │
│ 2022-11-22T23:31:38.8322182Z | RemoteAgent | TRACE | AgentHub received for 80, id 169, size 121                                                                            │
│ 2022-11-22T23:31:38.8341903Z | RemoteAgent | TRACE | IsProbeRequest: nslookup output : Server:        172.20.0.10\nAddress:    172.20.0.10:53\n\n131.92.0.10.in-addr.arpa  │
│ 2022-11-22T23:31:39.8327172Z | RemoteAgent | TRACE | ReversePortForwardConnector.HookupStreamData ReadAsync 169 returns 0 bytes.                                           │
│ 2022-11-22T23:31:39.8328106Z | RemoteAgent | TRACE | ReversePortForwardConnector.StartReceiveDataAsync finishes stream 169.                                                │
│ 2022-11-22T23:31:39.8328620Z | RemoteAgent | TRACE | PortForwardConnector.Disconnect 169                                                                                   │
│ 2022-11-22T23:31:39.8329116Z | RemoteAgent | TRACE | AgentHub closed for 80, id 169                                                                                        │
│ 2022-11-22T23:31:48.8316998Z | RemoteAgent | TRACE | ReversePortForwardConnector on port 80 accepted incoming request as stream 170.                                       │
│ 2022-11-22T23:31:48.8319022Z | RemoteAgent | TRACE | AgentHub connnected for 80, id 170                                                                                    │
│ 2022-11-22T23:31:48.8320072Z | RemoteAgent | TRACE | ReversePortForwardConnector.HookupStreamData ReadAsync 170 returns 121 bytes.                                         │
│ 2022-11-22T23:31:48.8320688Z | RemoteAgent | TRACE | AgentHub received for 80, id 170, size 121                                                                            │
│ 2022-11-22T23:31:48.8340228Z | RemoteAgent | TRACE | IsProbeRequest: nslookup output : Server:        172.20.0.10\nAddress:    172.20.0.10:53\n\n131.92.0.10.in-addr.arpa  │
│ 2022-11-22T23:31:49.8318578Z | RemoteAgent | TRACE | ReversePortForwardConnector.HookupStreamData ReadAsync 170 returns 0 bytes.                                           │
│ 2022-11-22T23:31:49.8319780Z | RemoteAgent | TRACE | ReversePortForwardConnector.StartReceiveDataAsync finishes stream 170.                                                │
│ 2022-11-22T23:31:49.8320468Z | RemoteAgent | TRACE | PortForwardConnector.Disconnect 170                                                                                   │
│ 2022-11-22T23:31:49.8321106Z | RemoteAgent | TRACE | AgentHub closed for 80, id 170                                                                                        │
│ 2022-11-22T23:31:58.8322014Z | RemoteAgent | TRACE | ReversePortForwardConnector on port 80 accepted incoming request as stream 171.                                       │
│ 2022-11-22T23:31:58.8324079Z | RemoteAgent | TRACE | AgentHub connnected for 80, id 171                                                                                    │
│ 2022-11-22T23:31:58.8325376Z | RemoteAgent | TRACE | ReversePortForwardConnector.HookupStreamData ReadAsync 171 returns 121 bytes.                                         │
│ 2022-11-22T23:31:58.8326184Z | RemoteAgent | TRACE | AgentHub received for 80, id 171, size 121                                                                            │
│ 2022-11-22T23:31:58.8354121Z | RemoteAgent | TRACE | IsProbeRequest: nslookup output : Server:        172.20.0.10\nAddress:    172.20.0.10:53\n\n131.92.0.10.in-addr.arpa  │
│ 2022-11-22T23:31:59.8321283Z | RemoteAgent | TRACE | ReversePortForwardConnector.HookupStreamData ReadAsync 171 returns 0 bytes.                                           │
│ 2022-11-22T23:31:59.8322305Z | RemoteAgent | TRACE | ReversePortForwardConnector.StartReceiveDataAsync finishes stream 171.                                                │
│ 2022-11-22T23:31:59.8323001Z | RemoteAgent | TRACE | PortForwardConnector.Disconnect 171                                                                                   │
│ 2022-11-22T23:31:59.8323626Z | RemoteAgent | TRACE | AgentHub closed for 80, id 171
```

**Mention the platform you are using**
vscode

**To Reproduce**
1. Configure Bridge to Kubernetes in VsCode and start tunneling traffic from service to local port
2. Run `kubectl get pods` to see that there are 2 pods for the corresponding service, because new pod (which has remote agent) is not ready
3. Run `kubectl describe pod {pod name}` to see that it fails readiness probe

**Expected behavior**
Pod with traffic agent starts successfully and starts to redirect traffic to local machine

**Desktop (please complete the following information):**
 - OS: MacOs Ventura 13.01
 - Browser: Chrome

Eneuman:
This PR disables probes from beeing cloned to the remote agent pod when the remote agent is deployed.
Existing probes are restored when EndpointManager disconnects.

You can enable cloning of probes by adding the following line to the KubernetesLocalProcessConfig.yaml file.

```
enableFeatures:
  - Probes  

```
fixes #114
fixes #111


Eneuman:
@elenavillamil I just noticed that the existing code in GetStatefulSetJsonPatch doesn't handle StartupProbes. Do you want me to fix this in this PR as well?

src/library/Connect/KubernetesRemoteEnvironmentManager.cs
```diff
@@ -1197,6 +1229,27 @@ private async Task<bool> _WaitForDeploymentChangedAsync(string namespaceName, st
...
+                {
+                    patch.Remove(d => d.Spec.Template.Spec.Containers[containerIndex].LivenessProbe);
+                    reversePatch.Add(d => d.Spec.Template.Spec.Containers[containerIndex].LivenessProbe, statefulSet.Spec.Template.Spec.Containers[containerIndex].LivenessProbe);
+                }
+
+                if (statefulSet.Spec.Template.Spec.Containers[containerIndex].ReadinessProbe != null)
+                {
+                    patch.Remove(d => d.Spec.Template.Spec.Containers[containerIndex].ReadinessProbe);
+                    reversePatch.Add(d => d.Spec.Template.Spec.Containers[containerIndex].ReadinessProbe, statefulSet.Spec.Template.Spec.Containers[containerIndex].ReadinessProbe);
+                }
```
elenavillamil, line 1244: Believe it or not, this function already has this logic around line 1172 :) I had done some research around this for this issue https://github.com/Azure/Bridge-To-Kubernetes/issues/114 and pasted some of the dev related findings on it. For some reason this code path which tights to _PatchStatefulSetAsync has probes removed but the other code paths dont. I will leave more details in the overall review comment, but for this part I would suggest removing old code that removes liviness and readiness probe in this function and keep this new block which incluses startupprobe as well.

src/library/Connect/KubernetesRemoteEnvironmentManager.cs
```diff
@@ -114,22 +114,22 @@ internal class KubernetesRemoteEnvironmentManager : IRemoteEnvironmentManager
                         break;
                     // Clone pod spec and replace container with agent
                     case RemoteAgentHostingMode.NewPodWithContext:
-                        podAndContainer = await this._ClonePodAsync(remoteContainerConnectionDetails, cancellationToken);
+                        podAndContainer = await this._ClonePodAsync(remoteContainerConnectionDetails, localProcessConfig, cancellationToken);
```
elenavillamil, line 117: This code came to us with minimal UTs. So as we add new features / fix issues, we are trying to add UTs with the change that covers the parts being changed. It would be great if you could add one test file for this file in library.tests and add UTs to cover the 3 switches on this method.

elenavillamil (commented):
Thank you again for your contribution :) Regarding the comment on _GetStatefulSetJsonPathch, I am trying to think what is best way forward for backwards compatibility. I can think of two options:

1. Remove the flag and bridge always removes probes for all code paths by default. A bit scared as to why old team did not do that from the begining.
2. Keep flag and make it default true. So even if user doesnt add it we will remove it, but it will be user option to keep probes it if they want to.

I am leaning towards option 2 since it adds more flexibility for not that much more work. All we would have to do is add a new section (DisabledFeatures) instead of adding this new setting under EnabledFeatures on the configuration yaml file.

elenavillamil:
> @elenavillamil I just noticed that the existing code in GetStatefulSetJsonPatch doesn't handle StartupProbes. Do you want me to fix this in this PR as well?

I think I was in the middle of reviewing this as you commented this :) I just posted the review, let me know if you have questions after reading the comments.

Eneuman:
> Thank you again for your contribution :) Regarding the comment on _GetStatefulSetJsonPathch, I am trying to think what is best way forward for backwards compatibility. I can think of two options:
> 
> 1. Remove the flag and bridge always removes probes for all code paths by default. A bit scared as to why old team did not do that from the begining.
> 2. Keep flag and make it default true. So even if user doesnt add it we will remove it, but it will be user option to keep probes it if they want to.
> 
> I am leaning towards option 2 since it adds more flexibility for not that much more work. All we would have to do is add a new section (DisabledFeatures) instead of adding this new setting under EnabledFeatures on the configuration yaml file.

@elenavillamil Sure, sound good to me, but shouldn't probes be a "enable" feature in that case?
Like this?

version: 0.1
enableFeatures:
  - ManagedIdentity
  - Probes

elenavillamil:
> > Thank you again for your contribution :) Regarding the comment on _GetStatefulSetJsonPathch, I am trying to think what is best way forward for backwards compatibility. I can think of two options:
> > 
> > 1. Remove the flag and bridge always removes probes for all code paths by default. A bit scared as to why old team did not do that from the begining.
> > 2. Keep flag and make it default true. So even if user doesnt add it we will remove it, but it will be user option to keep probes it if they want to.
> > 
> > I am leaning towards option 2 since it adds more flexibility for not that much more work. All we would have to do is add a new section (DisabledFeatures) instead of adding this new setting under EnabledFeatures on the configuration yaml file.
> 
> @elenavillamil Sure, sound good to me, but shouldn't probes be a "enable" feature in that case? Like this?
> 
> ```
> version: 0.1
> enableFeatures:
>   - ManagedIdentity
>   - Probes
> ```

This is another way of thinking of it. I like it. If Probes is enabled then Bridge keeps the configuration for the probes. Else it will remove them. Once this is checked in and released I will ask our PM to update the public documentation.

sabbour:
Looks good to me. Are there any scenarios where a user would want to have the probes enabled by default in the remote agent that would break with this release?

src/library/Connect/KubernetesRemoteEnvironmentManager.cs
Eneuman: Done :)

How about this code part, do you know why StartupProbes are not set to null here?
https://github.com/Azure/Bridge-To-Kubernetes/blob/59c7d82d03668b322d9270f497e31ac7b6521333/src/library/Connect/KubernetesRemoteEnvironmentManager.cs#L486

src/library/Connect/KubernetesRemoteEnvironmentManager.cs
Eneuman: Yeah I was going to add Tests to this but I need advice on how to proceed here.

Todays code needs to be refactored if we want to add some meningfull tests for this function.
It calls alot of private functions and I havn't figured out how to test this in the current mock framework.
 
Do you want me to refactor it in this PR or create a seperate PR for the refactoring and adding tests?

src/library/Connect/KubernetesRemoteEnvironmentManager.cs
elenavillamil: no, unfortunately I dont have the historical reasons on this part. However, I think it should be save to delete startup probes as well.

elenavillamil:
Its also worth noting that B2K code has handle to return 200 for probes [here](https://github.com/Azure/Bridge-To-Kubernetes/blob/1d363ab57a95556aae12ec27eb9825a0855261e6/src/devhostagent/PortForward/ReversePortForwardConnector.cs#L146). Personally, I think just ignoring the probes is a simpler and less error prone solution than trying to intercept probe requests and return 200. But wanted to drop this here for discussion.

src/library/Connect/KubernetesRemoteEnvironmentManager.cs
Eneuman: Ok, I have removed the startup probe aswell

elenavillamil:
> > Its also worth noting that B2K code has handle to return 200 for probes [here](https://github.com/Azure/Bridge-To-Kubernetes/blob/1d363ab57a95556aae12ec27eb9825a0855261e6/src/devhostagent/PortForward/ReversePortForwardConnector.cs#L146). Personally, I think just ignoring the probes is a simpler and less error prone solution than trying to intercept probe requests and return 200. But wanted to drop this here for discussion. @Eneuman and @hsubramanianaks as FYI for thoughts :)
> 
> @elenavillamil @hsubramanianaks My experience is that the todays handling of probe requests sometime works but mostly not. Now that I disabled the probes it feels more solid. If the pod is always replaced with the devagent then I cannot think of a scenario where probes are needed.
> 
> Maby we should add a condition to the code you referenced to @elenavillamil so It also gets disabled when probes are not enabled?

Yes, I think so. I like the idea of associating that code with the flag we are introducing with this change. This will remove forwarding complexity in the general case and it will make it cleaner if we every want to drop that flag and code path.

Eneuman:
> > > Its also worth noting that B2K code has handle to return 200 for probes [here](https://github.com/Azure/Bridge-To-Kubernetes/blob/1d363ab57a95556aae12ec27eb9825a0855261e6/src/devhostagent/PortForward/ReversePortForwardConnector.cs#L146). Personally, I think just ignoring the probes is a simpler and less error prone solution than trying to intercept probe requests and return 200. But wanted to drop this here for discussion. @Eneuman and @hsubramanianaks as FYI for thoughts :)
> > 
> > 
> > @elenavillamil @hsubramanianaks My experience is that the todays handling of probe requests sometime works but mostly not. Now that I disabled the probes it feels more solid. If the pod is always replaced with the devagent then I cannot think of a scenario where probes are needed.
> > Maby we should add a condition to the code you referenced to @elenavillamil so It also gets disabled when probes are not enabled?
> 
> Yes, I think so. I like the idea of associating that code with the flag we are introducing with this change. This will remove forwarding complexity in the general case and it will make it cleaner if we every want to drop that flag and code path.

I took a second look about this and since the remoteContainerConnectionDetails won't collect the probes from the spec, the probe information will not be send to AgentExecutorHub in the dev agent, so the RunReversePortForward command will not have any probes to check IsProbeRequest for :)

Maby I should just add a comment about this in the code?

src/library/Connect/KubernetesRemoteEnvironmentManager.cs
elenavillamil: @Eneuman, I was hoping that mocking the variables used for the switch statements was enough. If this is not enough, then yes we can leave it to a later refactor PR.

elenavillamil (approved):
@Eneuman Regarding DAPR, I have checked out [our manual testing for DAPR](https://github.com/Azure/Bridge-To-Kubernetes/blob/main/tests/manual/BridgeToKubernetes.md#dapr) and the sample application we use doesnt have probes. Thus, I am pretty sure probes are not a requirement for DAPR. 

Yes, please add comment so in future we know that may be code we can drop if after deploying this we dont see any usages of the Probes flag.

I am sending this to our testing team next and I will keep this updated with any findings.

elenavillamil:
Our testing team found that when running VSIX with this change restore job cannot complete. This is using stats-api from todo-app in samples folder. I can reproduce locally as well. If I add the Probes under enableFeatures in the yaml file, then it runs fine. I have to look at restore job logic to better understand what is happening. I tried to get logs but I couldnt get anything useful, pod went into error state and restarted. The logs I got before didnt have much. I will keep trying to get logs of what causes pod to go into error state. We may have this in our telemetry.
Screenshot 2023-02-22 at 8 46 24 PM

Eneuman:
@elenavillamil Could you ask them to do a descibe on the pod and check the events?

Eneuman:
I think I now know what's going on. Will try and fix this tomorrow.

elenavillamil:
> I think I now know what's going on. Will try and fix this tomorrow.

@Eneuman Sounds good! Let us know if you need help.

elenavillamil:
> > I think I now know what's going on. Will try and fix this tomorrow.
> 
> @Eneuman Sounds good! Let us know if you need help.

Hi @Eneuman, happy to take over this pull request if you got busy. Just let us know, I dont want to over step :)

Eneuman:
@elenavillamil 
Please have a look if you got the time :)

I think the problem might have something to do with using json patch together with removing the probes but I can be wrong here.

Eneuman:
@elenavillamil I did some testing yesterday and I can't reproduce this exception on my machine any more :(

I tried using the latest extensions in both VS Code and VS Pro, both set to my local build of DSC.

elenavillamil:
@Eneuman, yes I was testing this Friday and I could not repo either. I sent it off this morning to our testing team for final validation. The only thing I noticed is, if you look at restore logs when restore is happening is below warning. Which just indicates restore is being quicked from outside restore pod which is okay.

2023-03-20T17:12:05.2072004Z | RestorationJob | WARNG | Found 2 pods for deployment todo-app/stats-api but expected 1
2023-03-20T17:12:05.2087181Z | RestorationJob | TRACE | Couldn't get agent endpoint

hsubramanianaks (commented):
Thanks @Eneuman for these changes and @elenavillamil for reviewing it. Only thing I would call out is either in the README or document update, please have something similar ```By default bridge would remove the probes (liveness, readiness and startup) for the remote agent pod deployed in cluster, to enable them please use EnableFeatureProbes to true```.  Because at first glance I thought this is disabling probes (PR description says that) but if this flag is not set or set to any value other than true, bridge would disable probes. Thanks.

elenavillamil:
> Thanks @Eneuman for these changes and @elenavillamil for reviewing it. Only thing I would call out is either in the README or document update, please have something similar `By default bridge would remove the probes (liveness, readiness and startup) for the remote agent pod deployed in cluster, to enable them please use EnableFeatureProbes to true`. Because at first glance I thought this is disabling probes (PR description says that) and got confused. But if this flag is not set or set to any value other than true, bridge would disable probes. Thanks.

Correct, plan is to update public docs once this is released. @sabbour / @qpetraroia as FYI. I will reach out once this is released for this doc update.

PR #168, opened 2023-02-17
fix for enotfound getaddrinfo

hsubramanianaks:
This PR is to fix followin ENOTFOUND error while debugging samples/frontend app

image

Tests to confirm with screenshots, WIP

- Windows 
- Linux
- Mac


samples/todo-app/frontend/package.json
```diff
@@ -15,7 +15,7 @@
         "react": "^17.0.2",
         "react-dom": "^17.0.2",
         "request": "^2.88.2",
-        "servicebus": "^2.3.0",
+        "@servicebus/rabbitbus": "^1.0.14",
```
Tatsinnit, line 18: 💡☕️ This looks good, just mentioning for the archival reasons and folks might be curious:

* THe actual package which is old package was last released 4 years back : https://www.npmjs.com/package/servicebus?activeTab=versions 

#### How `rabbitbus` is a better choice?

Because it is recommended by the authors of the `servicebus` https://github.com/mateodelnorte/servicebus/issues/179

samples/todo-app/frontend/server.js
```diff
@@ -24,7 +24,7 @@ app.prepare().then(() => {
         });
     });
 
-    var bus = serviceBus.bus({ url: process.env.STATS_QUEUE_URI });
+    var bus = serviceBus.bus({ url: "amqp://"+process.env.STATS_QUEUE_SERVICE_HOST+":"+process.env.STATS_QUEUE_SERVICE_PORT_TCP });
```
Tatsinnit, line 27: 💡☕️ Might be worth to just add debug output for how and why `process.env.STATS_QUEUE_URI` to this `"amqp://"+process.env.STATS_QUEUE_SERVICE_HOST+":"+process.env.STATS_QUEUE_SERVICE_PORT_TCP` is valid.

* Also, super Nit -  we could do string literalise it like: `Foo ${object}` 

```.ts
var bus = serviceBus.bus({ url: `amqp://${process.env.STATS_QUEUE_SERVICE_HOST}:${process.env.STATS_QUEUE_SERVICE_PORT_TCP}` });
```

Tatsinnit (approved):
💡☕️❤️ Thank you so much for the ping and adding me to this review - left just couple of comments and observations, also rabbitmq use is already test manually by you I guess. Rest looks good, thanks heaps.

samples/todo-app/frontend/package.json
hsubramanianaks: Tats, Thank you. You want me to add a comment why this dependency is replaced ?

samples/todo-app/frontend/server.js
hsubramanianaks: Tats , like the logger statement ?

samples/todo-app/frontend/package.json
Tatsinnit: ☕️✅ Sounds good, if I May recommend- leave this comment unresolved, for any future reader. Commit history should have it. Or might be to be succinct ass commented link of above tool recommendations as a reference.

elenavillamil (dismissed):
Looks good. Left minor comment. Verified on Mac as well.

samples/todo-app/frontend/server.js
Tatsinnit: ☕️💭💡nice, re: ampq - might be needed for rabbitmq 🤷🏽‍♂️ 

☕️ Also, the nit above is just to use the goodness of  template literals (back tick) with interpolated expressions  rather than string concatenation, all your choice whatever you feel like. 

Thanks heaps 🥷🥷

hsubramanianaks: Thanks for these good suggestions. Sample value of process.env.STATS_QUEUE_URI would be ```amqp://stats-queue``` which includes the ampq, I think that's why Elena would have suggested.  @elenavillamil It didn't work with stats_queue_uri for some reason. That's why I switched to service_host. Please see below for test results..

image

image

image

hsubramanianaks: @Tatsinnit I did read about interpolated template literals vs concatenating string, both seems similar. Only benefit would readability. Updated the code accordingly. Thanks for the suggestion.

samples/todo-app/frontend/package.json
hsubramanianaks: I am leaving this comment unresolved for historical purposes.

Tatsinnit (approved):

elenavillamil (approved):

samples/todo-app/frontend/package.json
hsubramanianaks: @Tatsinnit resolving this comment and we can always come back and look at resolved one's. Thank you.

Issue #165, opened 2023-02-13
Named port environment variables not being created


Strayfe:
I am raising this as a new issue as I am not sure I am going to get a response to https://github.com/Azure/Bridge-To-Kubernetes/issues/64

There seems to be a few discrepancies between the way that Kubernetes and Bridge allocate ports from a service that contains multiple named ports.

There are environment variables created by Kubernetes that reference ports by the **name** metadata, if it exists, instead of by port which are allocated arbitrarily when using Bridge as the OP described.

Say your microservice has a dependency on a service you have no control over, let's use the bitnami RabbitMQ helm chart  as an example, which exposes the following service (trimmed for brevity): 

```
apiVersion: v1
kind: Service
metadata:
  name: rabbitmq
  namespace: rabbitmq
spec:
  ports:
  - name: amqp
    port: 5672
    protocol: TCP
    targetPort: amqp
  - name: epmd
    port: 4369
    protocol: TCP
    targetPort: epmd
  - name: dist
    port: 25672
    protocol: TCP
    targetPort: dist
  - name: http-stats
    port: 15672
    protocol: TCP
    targetPort: stats
``` 

This service has 4 ports described, all with `name` metadata so the ports can be referenced by name in other resources like Ingresses, etc.

An example of this would look like: 

```
apiVersion: networking.k8s.io/v1
kind: Ingress
...
spec:
  rules:
  - host: rabbitmq.local
    http:
      paths:
      - backend:
          service:
            name: rabbitmq
            port:
              name: http-stats        <-- port referenced by name, not port number
```

Now, when we look at one of our RabbitMQ pods in a Kubernetes cluster, we can see environment variables that have been created to reflect each of these ports by **name** (IP addresses obfuscated to 0.0.0.0):

```
<<K9s-Shell>> Pod: rabbitmq/rabbitmq-0 | Container: rabbitmq 
I have no name!@rabbitmq-0:/$ printenv | grep RABBITMQ_
RABBITMQ_PORT_5672_TCP_ADDR=0.0.0.0
RABBITMQ_PORT_15672_TCP_ADDR=0.0.0.0
RABBITMQ_PORT_5672_TCP_PORT=5672              <-- (c)
RABBITMQ_PORT_4369_TCP_PORT=4369
RABBITMQ_SERVICE_PORT_AMQP=5672               <-- (d)
RABBITMQ_PORT_25672_TCP_ADDR=0.0.0.0
[...]

PR #172, opened 2023-02-22
Named Port Environment Variables

Strayfe:
Kubernetes used to have an environment variable of `{serviceName}_SERVICE_PORT_{protocol}`, however, this has since been replaced with a better mechanism for tracking services that contain multiple port mappings using name metadata to keep the ports unambiguous.

> When using multiple ports for a Service, you must give all of your ports names so that these are unambiguous.
https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services

Issue: https://github.com/Azure/Bridge-To-Kubernetes/issues/165

Whilst I have refactored the unit tests for the library code and they are passing, I have not yet looked into actually testing this out in VSCode, I am assuming there are instructions to do this somewhere or can someone try it for me?


Strayfe:

PR #173, opened 2023-02-22
Multi-Port Services: Named Port Environment Variables

Strayfe:
Kubernetes used to have an environment variable of `{serviceName}_SERVICE_PORT_{protocol}`, however, this has since been replaced with a better mechanism for tracking services that contain multiple port mappings using name metadata to keep the ports unambiguous.

> When using multiple ports for a Service, you must give all of your ports names so that these are unambiguous.
https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services

Issue: https://github.com/Azure/Bridge-To-Kubernetes/issues/165


src/library/Connect/LocalEnvironmentManager.cs
```diff
@@ -421,67 +421,69 @@ private void _ReportProgress(EventLevel eventLevel, string message, params objec
...
+                    .First()
+                    .Replace("-", "_")
+                    .Replace(".", "_");
+
+                var host = _useKubernetesServiceEnvironmentVariables || string.Equals(endpoint.DnsName, DAPR, StringComparison.OrdinalIgnoreCase)
+                    ? endpoint.LocalIP.ToString()
+                    : endpoint.DnsName;
+
+                if (string.Equals(serviceName, "KUBERNETES", StringComparison.OrdinalIgnoreCase))
+                    host = _kubernetesClient.HostName;
```
elenavillamil, line 450: nit: can you keep {} for styling with the rest of the code.

            if (serviceName == "KUBERNETES")
            {
                // reset KUBERNETES_SERVICE_HOST to cluster name.
                host = this._kubernetesClient.HostName;
            }

src/library/Connect/LocalEnvironmentManager.cs
```diff
@@ -421,67 +421,69 @@ private void _ReportProgress(EventLevel eventLevel, string message, params objec
...
-        /// <summary>
-        /// Add Kubernetes environment variables to existing list of environment variables
-        /// </summary>
-        private void _AddKubernetesServiceEnv(IDictionary<string, string> envVariables, string dnsName, string host, string protocol, int port)
-        {
-            dnsName = dnsName.ToUpperInvariant();
-            // because we are using dns name instead of service we have to retrieve it by splitting when needed
-            // If this ever cause issues we should consider larger refactor where we add serviceName member varaible to EndpointInfo class.
-            string serviceName = dnsName.Split(".").First();
+                    var protocolUpper = portPair.Protocol.ToUpperInvariant();
```
elenavillamil, line 470: Before there was a check for this being empty and if so it would set it to tcp. Can we keep the check?

src/library/Connect/LocalEnvironmentManager.cs
```diff
@@ -421,67 +421,69 @@ private void _ReportProgress(EventLevel eventLevel, string message, params objec
...
+                    .Replace(".", "_");
+
+                var host = _useKubernetesServiceEnvironmentVariables || string.Equals(endpoint.DnsName, DAPR, StringComparison.OrdinalIgnoreCase)
+                    ? endpoint.LocalIP.ToString()
+                    : endpoint.DnsName;
+
+                if (string.Equals(serviceName, "KUBERNETES", StringComparison.OrdinalIgnoreCase))
+                    host = _kubernetesClient.HostName;
+
+                var unnamedPort = _useKubernetesServiceEnvironmentVariables || string.Equals(endpoint.DnsName, DAPR, StringComparison.OrdinalIgnoreCase)
```
elenavillamil, line 452: nit: Can we have comment explaining why this unnamedPort is needed and how it is different from port in line 466? or maybe just move comment in 459 here?

elenavillamil (commented):
Hi @Strayfe thank you for your PR! It looks good, I just left some minor comments. Thank you for keeping UTs up to date as well! 

Regarding end to end, did you test this locally yourself? It is okay if you didn't, i can do the basic test for you. I just want to have some end to end testing done before sending to our testing team.

src/library/Connect/LocalEnvironmentManager.cs
Strayfe: Absolutely, will patch it in the upcoming commit.

src/library/Connect/LocalEnvironmentManager.cs
Strayfe: Ah yes, I can see no existing code paths currently set this property directly and both constructors default it to the constant `KubernetesConstants.Protocols.Tcp` with no available empty constructor to allow for object initialization but it could be overwritten after object creation.

I will write a defensive check for this specific scenario here but I am concerned this would also need to be replicated in every place where the value of this property needs to be defaulted.

In the interest of DRY, could there be an argument for adding a default/empty constructor to allow object initialization and defaulting the property so this would become the ubiquitous default for this property, both steps are required to avoid a code smell:

```c# 
public PortPair() 
{
}

public string Protocol { get; set; } = KubernetesConstants.Protocols.Tcp;
```

src/library/Connect/LocalEnvironmentManager.cs
Strayfe: Ah yes!

> 2. The SERVICE_PORT variable should map to first entry in config not last.

This environment variable only exists once per service regardless of the number of ports and Kubernetes seems to allocate it to the first port in the YAML definition, not the last, so moving it outside the loop and allocating it specifically once at the start of the process should fix this secondary issue.

I will add a comment to this effect to the code 😄

Strayfe:
> Hi @Strayfe thank you for your PR! It looks good, I just left some minor comments. Thank you for keeping UTs up to date as well!

You are very welcome, I actually found this to be quite a fun challenge, the code is clean and easy to understand and I enjoyed working on the UTs!

> Regarding end to end, did you test this locally yourself? It is okay if you didn't, i can do the basic test for you. I just want to have some end to end testing done before sending to our testing team.

Sadly, I was not able to figure out how to test this locally, however, I must admit I did not try very hard to look for resources to figure out how, past looking at the readme in the library project 😆, if you wouldn't mind pointing me in the direction of some materials to test this e2e I would be happy to give it a go, else I would appreciate it if you could see if it works!

elenavillamil:
> 

Thanks @Strayfe! Since this does not have changes to the docker images, just the CLI (what code structure has it has dsc folder), it should be pretty straight forward to give it a try. [This readme](https://github.com/Azure/Bridge-To-Kubernetes/tree/main/src/dsc#readme) has the details under building section, essentially you just build and publish the dsc folder. then you set environment variable BRIDGE_BUILD_PATH to the path of the publish output. then open VSCode from that window where you set the environment variable and you will be running bridge using your code. Let me know if you run into any issues. If it doesnt work an you need to debug, you can follow steps on Developing/Debugging on that section, you will be able to step through the code (you may need ot update launch.json per your cluster namespace and service and computer architecture).

elenavillamil (approved):
LGMT! Code looks good and testing team approved.

PR #175, opened 2023-02-24
fix case sensitivity and update base images for mariner

gambtho:
Updating images to use mariner versions - https://learn.microsoft.com/en-us/azure/aks/use-mariner


samples/ingress/database-api/Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM mcr.microsoft.com/dotnet/core/aspnet:3.1-buster-slim AS base
```
gambtho, line 1: need to test this manually

src/routingmanager/Dockerfile
```diff
@@ -20,13 +20,12 @@ ENV MINDARO_BUILD_NUMBER=${MindaroBuildNumber}
 RUN dotnet publish -c ${Configuration} -o /src/publish
 
 # Final container
-FROM mcr.microsoft.com/dotnet/aspnet:6.0-alpine3.16 as final
```
gambtho, line 23: this is slightly bigger, but not a huge difference

src/devhostagent/Dockerfile
```diff
@@ -20,7 +20,7 @@ ENV MINDARO_BUILD_NUMBER=${MindaroBuildNumber}
 RUN dotnet publish -c ${Configuration} -o /src/publish devhostAgent.csproj
 
 # Final container
-FROM mcr.microsoft.com/dotnet/aspnet:6.0-alpine3.16 as final
```
gambtho, line 23: this is significantly smaller and more secure

src/LocalAgent/Dockerfile
```diff
@@ -1,28 +1,31 @@
...
 WORKDIR /src
-COPY ["nuget.config", "."]
-COPY ["LocalAgent/localagent.csproj", "LocalAgent/"]
-COPY ["library/library.csproj", "library/"]
-COPY ["common/common.csproj", "common/"]
-COPY ["common.json/common.json.csproj", "common.json/"]
-RUN dotnet restore "LocalAgent/localagent.csproj"
+
+COPY ["/src/nuget.config", "."]
+COPY ["/src/LocalAgent/LocalAgent.csproj", "LocalAgent/"]
```
Tatsinnit, line 12: 💡 Observation/thoughts 🥷☕️: if we have `workingdir` already mentioned as `/src` then I think `COPY ...` will route through `/src` https://docs.docker.com/engine/reference/builder/#workdir

samples/ingress/database-api/Dockerfile
Tatsinnit, line 1: 💡 This is interesting, there was a `dotnet6` upgrade PR in last release, so is this something missed or `3.0` was in user for some special case. cc: @hsubramanianaks

src/LocalAgent/Dockerfile
```diff
@@ -1,28 +1,31 @@
...
+RUN dotnet build "LocalAgent.csproj" -c Release -o /app/build
 
 FROM build AS publish
-RUN dotnet publish "localagent.csproj" -c Release -o /app/publish
+RUN dotnet publish "LocalAgent.csproj" -c Release -o /app/publish
 
 FROM base AS final
-RUN apt-get update
-RUN apt-get install iptables
+RUN tdnf install -y \
```
Tatsinnit, line 27: 💡 I wonder if this is some auto tool reindenting and removing `update.`

Tatsinnit (dismissed):
❤️☕️ +1,  Thank you for adding this and for this PR, looks great, just added some observation, thanks heaps! (I don't quiet know how this multi-step docker files fit together but 100 % agree that we should manually test this)

samples/ingress/database-api/Dockerfile
hsubramanianaks: this is in samples/todo-app which we use extensively for testing. Actual docker images for devhost, restoration and routing manager were updated to dotnet 6. :)

build/setup-kubectl.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 
 DEFAULT_INSTALL_LOCATION="/app/kubectl/linux"
-DEFAULT_KUBECTL_VERSION="v1.21.2"
+DEFAULT_KUBECTL_VERSION="v1.25.5"
```
hsubramanianaks, line 5: This requires testing routing manager which uses this kubectl. I will send the request to Joyer to test it.

src/devhostagent.restorationjob/Dockerfile
```diff
@@ -20,15 +20,17 @@ ENV MINDARO_BUILD_NUMBER=${MindaroBuildNumber}
 RUN dotnet publish -c ${Configuration} -o /output devhostAgent.restorationjob.csproj
 
 # Final container
-FROM mcr.microsoft.com/dotnet/aspnet:6.0-alpine3.16 as final
+FROM mcr.microsoft.com/dotnet/aspnet:6.0-cbl-mariner2.0 as final
```
hsubramanianaks, line 23: Can we try distroless image ```FROM mcr.microsoft.com/dotnet/runtime-deps:6.0-cbl-mariner2.0-distroless as final ``` ? which is more performant and anyway we are installing the OS dependencies, tools manually using tndf package manager. This is just a suggestion if this causes any problems, we can always revert back.

hsubramanianaks (dismissed):
Just few suggestions, other than that everything looks good to me.

src/LocalAgent/Dockerfile
hsubramanianaks: I think the purpose here is to use tndf package manager which comes with CBL-Mariner which might not need update any more. But Tom can always correct me if  iam wrong. Thank you. :)

gambtho: yeah, mariner doesn't require update and uses tdnf instead of apt :)

src/devhostagent.restorationjob/Dockerfile
gambtho: distroless is missing the dependencies installed on 26-29, unless we don't need those?

hsubramanianaks: I thought distroless got those dependencies covered. if that's the case we can resolve this. :), Thanks.

src/LocalAgent/Dockerfile
gambtho: replaced with tdnf update command :)

hsubramanianaks (dismissed):
looks good and testing team didn't find any issue for a week. @gambtho Thanks for this PR.

deployment/settings/services/imagetag.setting
```diff
@@ -1,4 +1,4 @@
-MINDARO_DEVHOSTAGENT_TAG=1.1.0
-MINDARO_DEVHOSTAGENT_RESTORATIONJOB_TAG=1.2.0
-MINDARO_ROUTINGMANAGER_TAG=1.1.0
+MINDARO_DEVHOSTAGENT_TAG=1.2.0
```
hsubramanianaks, line 1: @gambtho updated this so that B2K picks up new images or else existing k8's cluster will reuse the images.

hsubramanianaks (approved):

Issue #147, opened 2023-01-15
Saved cloned pod is not cleaned from added env variables


Eneuman:
When a pod is cloned, a backup of the pod spec is saved and  b2k env variables are added to the agent pod.
If b2k craches, thoose added env variables are not removed and no check if beeing made if they already exists before adding them.

This leeds to this:
![image](https://user-images.githubusercontent.com/12099277/212539223-3cf2c23b-878e-4565-aa44-5840befec168.png)

I suggest adding code to clean the saved pod spec from any added env variables.
This will make sure that the env variable only exists once.

Code could be added here:
https://github.com/Azure/Bridge-To-Kubernetes/blob/991c6a393f31fa581258f598b9a2dd1410a0a9f6/src/library/Connect/KubernetesRemoteEnvironmentManager.cs#L632

Vidya2606:
Hi @Eneuman can you please provide the steps to reproduce the issue ? When the bridge crashes how are you restoring it?

Eneuman:
@Vidya2606 Sure
Have a pod that contain probes running in your AKS.
Start debugging you local service.
Keep looking at your pods using k9s or something like that. The restore agents should crash and restart after 12 failed pings. When it do I think it uses the saved patch document and applies that, hence creating multiple instances of the env variables each time it restarts.

I haven't confirmed that's the case but my guess is the calls to DeplyPatchV1Async is causing this to happen.

hsubramanianaks:
This is released with version 1.0.20230418.1. Closing the issue.

PR #178, opened 2023-03-03
Modified method to have one set of env variables and added UTs

Vidya2606:
This PR is related to https://github.com/Azure/Bridge-To-Kubernetes/issues/147

I have Modified the method to add only the new environment variables, the old one's will not be added.
Also added the respective UT's


src/library/Connect/KubernetesRemoteEnvironmentManager.cs
```diff
@@ -468,11 +468,9 @@ private V1Pod _GetClonedPodSpec(V1Pod pod, string containerName, string agentIma
                     c.Image = agentImage;
                     c.Command = new List<string>();
                     c.Args = new List<string>();
+                    
+                    // When creating new env variables, the previous env variables are not retained to ensure only one is present.
                     var newEnv = new List<V1EnvVar>();
-                    if (c.Env != null)
```
Tatsinnit, line 472: 💡☕️ Thank you for this PR and thanks @hsubramanianaks for the ping to take a look, like openAI has identified, please refer here: https://github.com/Azure/Bridge-To-Kubernetes/pull/178#issuecomment-1495294974 

> The main issue I see is that the code is not retaining the previous environment variables when creating new ones. This could lead to unexpected behavior and should be addressed. I suggest adding a line of code to retain the previous environment variables when creating new ones. This should ensure that only one set of environment variables is present.

So is it what we are knowing aiming for or might be missing some behaviour here?

Tatsinnit: Nice, That is great then ,so the change of behaviour is expected in that case? (Worth to document this as an archive if we merge this PR)

hsubramanianaks: Agreed, will leave these comments as it is and will resolve conversation.  sounds good ?

hsubramanianaks (approved):
Testing team confirmed that this is working for windows and linux. Waiting for Mac results.

PR #184, opened 2023-03-04
Removed dependency to Minimatch

Eneuman:
Minimatch is a NuGet package for matching text patterns and is being used when deciding if a volume should be downloaded or not.
The nuget package only supports .Net Standard 1.0 and has not been activily maintained since 2016.

When bulding B2K, the following warning occures in the output:
`warning NU1603: Minimatch 2.0.0 depends on NETStandard.Library (>= 1.0.0-rc2-23910) but NETStandard.Library 1.0.0-rc2-23910 was not found. An approximate best match of NETStandard.Library 1.5.0-rc2-24027 was resolved.`

This PR removes the dependency to that package.

I have added the tests I could think of but the B2K documentation isn't clear on what wildcard patterns it supports.
Please have a look and see if I missed any test patterns.


Eneuman:
> notice.txt

@elenavillamil Fixed!

elenavillamil (approved):
LGMT. I validated locally and check against combinations of download volume i know of. Testing team also ran validation.

hsubramanianaks (approved):
Looks good to me.

Issue #90, opened 2022-11-06
The JSON value could not be converted to k8s.Models.V1ServiceStatus


arkadiya-msft:
**Describe the bug**
When trying to start a debugging session in VS Code, I see following error in B2K VS Code output:

```
Failed to establish a connection. Error: Connect operation failed.
An unexpected error occurred: 'The JSON value could not be converted to k8s.Models.V1ServiceStatus. Path: $.status | LineNumber: 0 | BytePositionInLine: 67.'
```

If I try again, it works as expected.

**Mention the platform you are using**
VS Code v1.72.2 
Extension v1.0.120221031
MacOS 13.0

**To Reproduce**
1. Run and Debug
2. Click on Start Debugging for a predefined launch configuration for B2K
3. Put in your credentials when prompted

**Expected behavior**
Debugging works the first time, without need to try again after first attempt fails.

**Additional context**

Stacktrace from MindaroCli logs:

[...]

Issue #55, opened 2022-09-15
The JSON value could not be converted to k8s.Models.V1ServiceStatus


elenavillamil:
**Describe the bug**
Bridge to Kubernetes sometimes fails with error: "An unexpected error occurred: 'The JSON value could not be converted to k8s.Models.V1ServiceStatus". Restarting VSCode will help and fix this issue. From metrics this started happening around August 07th.

**Mention the platform you are using**
All (Linux, Windows, Mac). VSCode.

**To Reproduce**
It happens sporadically when you start debugging using bridge.

**Expected behavior**
This error should not happen.

ajaffie:
Not 100% on this but it seems to happen when the routingmanager deployment is being created. Simply retrying the command has been successful for me. This happens for CLI and Visual Studio as well.

elenavillamil:
This is happening during routing manager deployment inside function call to DeleteNamespacedServiceAsync which is inside CreateOrReplaceV1ServiceAsync

elenavillamil:
Fix for this got deployed today. I. will keep an eye on our internal metrics and re-open if this still reproduces.

PR #187, opened 2023-03-06
Add catch around Delete Service invocation

elenavillamil:
there is breaking changes between different version of kubernetes if client and server diverge enough. 1.21 and 1.22 DeleteService returns v1.Status (6.0 client sdk) while in 1.23, DeleteService returns v1.Service (7.0+ client sdk). [Related issue](https://github.com/kubernetes-client/csharp/issues/824)  with more details. We can control the client since it comes with bridge code (aka c# library), but we cannot control the server aka kubernetes version in the cluster. Thus exception handling is needed. 

This will solve #90 and #55

### DIFF
--- a/src/common/Kubernetes/KubernetesClient.cs
+++ b/src/common/Kubernetes/KubernetesClient.cs
@@ -401,6 +401,7 @@ namespace Microsoft.BridgeToKubernetes.Common.Kubernetes
                 {
                     try 
                     {
+                        _log.Warning("Initial CreateNamespacedServiceAsync failed, deleting namespace");
                         await RestClient.CoreV1.DeleteNamespacedServiceAsync(service.Metadata.Name, namespaceName);
                     }
                     catch (JsonException ex)
--- a/src/common/Kubernetes/KubernetesClient.cs
+++ b/src/common/Kubernetes/KubernetesClient.cs
@@ -12,6 +12,7 @@ using System.Net.Http;
 using System.Net.Sockets;
 using System.Net.WebSockets;
 using System.Text;
+using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
 using k8s;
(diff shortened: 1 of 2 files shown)


src/common/Kubernetes/KubernetesClient.cs
```diff
@@ -398,8 +399,21 @@ public async Task<V1Service> CreateOrReplaceV1ServiceAsync(string namespaceName,
                 }
                 catch (HttpOperationException e) when (e.Response.StatusCode == HttpStatusCode.Conflict)
                 {
-                    await RestClient.CoreV1.DeleteNamespacedServiceAsync(service.Metadata.Name, namespaceName);
+                    try 
+                    {
+                        _log.Warning("Initial CreateNamespacedServiceAsync failed, deleting namespace");
+                        await RestClient.CoreV1.DeleteNamespacedServiceAsync(service.Metadata.Name, namespaceName);
+                    }
+                    catch (JsonException ex)
```
hsubramanianaks, line 407: @elenavillamil Please add a UT for this, KubernetesClientTests.cs exists.

elenavillamil: do you know how to mock DeleteNamespacedServiceAsync throwing exception? I looked at adding UT when I worked on this but got stock on that part.

elenavillamil: Ah figured it out. there is a throw method :)

elenavillamil: ug our current code uses FakeItEasy to mock things, unfortunately it seems they dont support mocking of static methods, which both  CoreV1.CreateNamespacedServiceAsync and CoreV1.DeleteNamespacedServiceAsync are. My guess this is why this code path does not have UTs to begin with. I will look at option of using other mocking tools.

[Fake It Easy Related Issue](https://github.com/FakeItEasy/FakeItEasy/issues/90)

elenavillamil: As a workaround, I noticed our KubernetesClient class had its own wrappers around it, so I tried changing code to use the wrappers (instead of calling IcoreV1 methods direclty) and then mock the wrappers, but when I tried to do so it i says _kClient is not mockable. From googling only interface classes can be mocked, so I tried making _kClient of type IKubernetesclient, but this causes more test to fails and doesnt solve the issue because if using interface when invoking method being tested the interface does not have the implementation for it. It becomes a chicken and egg problem (cant use object because mocking utility only support interface, but cant use interface because it doesnt have implementation for the function being tested). Hence, dont think adding UT coverage to this methods is really in scope for this PR.

hsubramanianaks: I think we can add work item for refactoring UT's or add more UT's in our backlog for now and pick it up later.

hsubramanianaks (approved):
LGTM, we can merge once testing team confirms it is working, Thanks for this PR.

PR #188, opened 2023-03-06
Add signing validation to our official build pipeline

elenavillamil:
Currently signing verification happens on release nuget package pipeline. This change aim to add this check sooner on the release process.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -551,12 +551,12 @@ jobs:
         }
 
         Write-Host "##vso[task.setvariable variable=MicroBuild_NuPkgSigningEnabled;]true"
-  #- task: MicroBuildSigningPlugin@3
-  #  displayName: Install Signing Plugin
-  #  enabled: True
-  #  inputs:
-  #    signType: real
-  #    zipSources: false
+  - task: MicroBuildSigningPlugin@3
+    displayName: Install Signing Plugin
+    enabled: True
+    inputs:
+      signType: real
+      zipSources: false
   - task: MicroBuildLocalizationPlugin@3
     displayName: Install Localization Plugin
   - task: UseDotNet@2
@@ -658,13 +658,13 @@ jobs:
         New-Item -Path '$(Build.SourcesDirectory)/src/dsc/bin/${{ parameters.BuildConfiguration }}/net6.0/linux-x64/publish/kubectl/linux' -ItemType Directory
 
         curl https://dl.k8s.io/release/v1.21.2/bin/linux/amd64/kubectl  -OutFile $(Build.SourcesDirectory)/src/dsc/bin/${{ parameters.BuildConfiguration }}/net6.0/linux-x64/publish/kubectl/linux/kubectl
-  - task: VSBuild@1
-    displayName: Sign Files
-    enabled: True
-    inputs:
-      solution: src/client.signproj
-      msbuildArgs: /t:SignFiles /p:SignType=real
-      configuration: '${{ parameters.BuildConfiguration }} '
+  #- task: VSBuild@1
+  #  displayName: Sign Files
+  #  enabled: True
+  #  inputs:
+  #    solution: src/client.signproj
+  #    msbuildArgs: /t:SignFiles /p:SignType=real
+  #    configuration: '${{ parameters.BuildConfiguration }} '
   - task: CopyFiles@2
     displayName: Collect files for .zip (Windows)
     inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -819,7 +819,6 @@ jobs:
     displayName: Validate Signatures linux
     inputs:
       Path: '$(Agent.TempDirectory)/zip/linux'
-      continueOnError: false
   - task: ArchiveFiles@2
     displayName: Create .zip file (Windows)
     inputs:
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -472,6 +472,10 @@ jobs:
     inputs:
       PathtoPublish: '$(Build.ArtifactStagingDirectory)/nuget'
       ArtifactName: nuget
+  - task: CodeSign@1
+    displayName: Validate Signatures nuget
```
hsubramanianaks, line 476: do we want to add ```continueOnError: true``` here ? I saw it is not failing, is it by default true ? just making sure.

azure-pipelines.yml
```diff
@@ -874,4 +906,19 @@ jobs:
     inputs:
       PathtoPublish: $(Build.ArtifactStagingDirectory)/zip
       ArtifactName: zipv2
+
+  - task: SdtReport@2
+    inputs:
+      GdnExportAllTools: true
+      GdnExportHtmlFile: true
```
hsubramanianaks, line 913: Small Suggestion/Idea: Currently report is exported only for errors , do we want to add "warnings" as well? because warnings turn into errors sometimes. :)
```GdnExportGdnToolCredScanSeverity: Warning```

File: azure-pipelines.yml [hunk 3]
```diff
@@ -874,4 +906,19 @@ jobs:
...
+  - task: PublishSecurityAnalysisLogs@3
+    inputs:
+      ArtifactName: 'CodeAnalysisLogs'
+      ArtifactType: 'Container'
+      AllTools: true
+      ToolLogsNotFoundAction: 'Standard'
+  - task: PostAnalysis@2
```
hsubramanianaks, line 920: Only suggestion I would have is, I see the build fails but doesn't show me reason of failure. someone new would struggle to find why build fails.  can we check if there is way for the post analysis to report that code sign validation failed or since we use this only for code sign , we can add only code sign validation to SdtReport@2 instead of all tools. Just an idea/suggestion. Thank you. 

image

hsubramanianaks (dismissed):
Added few suggestions/ideas , looks good to me.

azure-pipelines.yml
```diff
@@ -516,6 +520,22 @@ jobs:
     enabled: True
     inputs:
       tags: 'BuildConfiguration: ${{ parameters.BuildConfiguration }}'
+
+  - task: SdtReport@2
+    inputs:
+      GdnExportAllTools: true
```
hsubramanianaks, line 526: Suggestion/idea here is:
```inputs:
       GdnExportAllTools: false
       GdnExportHtmlFile: true
       GdnExportOutputBaselineFile: myBaseline
       GdnExportOutputBaseline: myBaselinedResults
       GdnExportOutputSuppressionFile: mySuppressions
       GdnExportOutputSuppressionSet: mySuppressionSet
       GdnExportGdnToolCodesignValidationSeverity: Warning
```

baseline file format/schema can be found [here](https://eng.ms/docs/microsoft-security/security/azure-security/cloudai-security-fundamentals-engineering/security-integration/guardian-wiki/microsoft-guardian/general/resource-samples/baseline-file), it is easily readable.

azure-pipelines.yml
elenavillamil: A lot of this checks have been changed to not fail on error but instead put logs. This is why I added the guardian break task at the end. I can give this flag a try.

azure-pipelines.yml
elenavillamil: So i had decided to keep it simple because our use case if pretty simple and want to avoid complexity. For example we dont have any need for suppressions, so kept that out. Code sign validation severity should be error hence didnt update default of that setting. What do you see the baseline being used for?

azure-pipelines.yml
elenavillamil: Will it also fail on warning? We only want the pipeline to fail on error right?

hsubramanianaks: No it will export both Errors and Warnings in the report. It will break only on Error or Errors and Warnings which can be configured via PostAnalysis step 
```- task: PostAnalysis@2
  displayName: 'Post Analysis'
  inputs:
    GdnBreakPolicyMinSev: Error``` Values can be Error or Warnings here. It will break the build based on this value.

azure-pipelines.yml
hsubramanianaks: baseline is just a json which is human readable to see which tool failed and why it failed.

azure-pipelines.yml
elenavillamil: Tried this glad now and it didnt work.

azure-pipelines.yml
hsubramanianaks: Needs more reading about guardian break, currently we are trying to keep this simple.

azure-pipelines.yml
elenavillamil: Tried some of these flags, but postanalysis still didnt include error logs in that task itself.

azure-pipelines.yml
elenavillamil: tested this flags, it doesnt change what shows up on logs of post analysis. Keeping it as error for now.

hsubramanianaks (approved):
LGTM

Issue #167, opened 2023-02-16
Services without cluster IP do not get forwarded


giulianob:
**Describe the bug**

I have a service named `metadata` which does not have a cluster IP assigned but has an endpoint. The service does not get added to my hosts file when I use `dsc`. 

The list of services I have:
> k get services
```
NAME                                TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)             AGE
metadata                            ClusterIP   None           <none>        7300/TCP            11m
frontend                            ClusterIP   10.0.175.119   <none>        80/TCP              19h
```

If I forward the `frontend` service then you can see `metadata` does not get added to my hosts file:
> dsc connect --service frontend
```
Waiting for 'frontend-6bf4d6f477-rjbt8' in namespace 'gbarberi2' to reach running state...
Deployment 'gbarberi2/frontend' patched to run agent.
Remote agent deployed in container 'frontend' in pod 'frontend-6bf4d6f477-rjbt8'.
Preparing to run Bridge To Kubernetes configured as pod gbarberi2/frontend-6bf4d6f477-rjbt8 ...
Connection established.
Hosts file updated.
Service 'frontend' is available on 127.1.1.17:80.
Container port 80 is available at localhost:null.
```

The `metadata` service does have an endpoint:
> k describe service metadata
```
Name:              metadata
Namespace:         gbarberi2
Labels:            app.kubernetes.io/managed-by=Helm
Annotations:       meta.helm.sh/release-name: metadata
                   meta.helm.sh/release-namespace: gbarberi2
Selector:          app=metadata
Type:              ClusterIP
IP Family Policy:  SingleStack
IP Families:       IPv4
IP:                None
IPs:               None
Port:              grpc  7300/TCP
TargetPort:        7300/TCP
Endpoints:         10.240.10.189:7300
Session Affinity:  None
Events:            <none>
```

I can port forward the `metadata` service with `kubectl` 
> k port-forward service/metadata 7300:7300
```
Forwarding from 127.0.0.1:7300 -> 7300
Forwarding from [::1]:7300 -> 7300
```

I can also exec into the service in my cluster and ping `metadata` so I would expect it to be also be available locally:
> k exec --stdin --tty frontend-57b9f954df-2kpr5 -- /bin/bash
```
root@frontend-57b9f954df-2kpr5:/App# ping metadata
PING metadata.gbarberi2.svc.cluster.local (10.240.10.189) 56(84) bytes of data.
[...]

Issue #155, opened 2023-01-30
Unable to resolve DNS names for certain Cluster-IP services 


d6teo:
**Describe the bug**
**Issue started after Visual Studio's extension upgrade to version 3.1.20230118.1.**
Debugged services are not able to resolve DNS names to IP addresses for deployments that expose both a headless service and a Cluster-IP service.
(does not seem to occur when a deployment exposes a single Cluster-IP service)

Examples: 
Redis (bitnami's Redis chart, StackExchangeRedis package used in debugged service): No such host is known 
RabbitMQ (bitnami's RebbitMQ chart, RabbitMQ.Client package used in debugged service): No ip address could be resolved for (DNS name of service)
Also occurs when trying to follow the todoapi tutorial (as described in #150) 

**Mention the platform you are using**
Microsoft Visual Studio Professional 2022, Version 17.4.3 
Bridge to Kubernetes 2022 Version 3.1.20230118.1
Minikube Version: v1.23.2 running on HyperV

**Expected behavior**
Debugged services should be able to resolve DNS names to IP addresses for all deployed services.

**Screenshots**
Deployed services:
![image](https://user-images.githubusercontent.com/22095087/215506945-b758ab97-5de9-448d-9c4b-5b0efa33729d.png)
BTK8s setup for service 'activityapi':
![image](https://user-images.githubusercontent.com/22095087/215507310-4fc33f44-614c-4397-9ef1-c5aabb4274ae.png)
Exception when debugged 'activityapi' services tries to resolve IP for service 'eventbus-development-rabbitmq' (it is however able to resolve other services whose deployments do not expose a headless service, e.g. 'periodicdataapi'):
![image](https://user-images.githubusercontent.com/22095087/215507780-c9ca1f2a-5f58-4fb7-840c-5067ddec7e55.png)

**Desktop (please complete the following information):**
 - OS: Windows 10 Enterprise Build 19045.2486

elenavillamil:
@d6teo thank you for raising this issue. You mentioned "Also occurs when trying to follow the todoapi tutorial (as described in https://github.com/Azure/Bridge-To-Kubernetes/issues/150)", but todo app tutorial doesnt have any headless service. Do you have a sample application we could use to debug/reproduce on our end? Its okay if you dont, but it would help if you do.

[...]

PR #190, opened 2023-03-08
Use endpoint dns if hostname not specified for the address

elenavillamil:
This Pr address incorrect assumption done on [this PR](https://github.com/Azure/Bridge-To-Kubernetes/pull/78/files#diff-c415242ad7d5ebbf107822a6ba3ee5f15ccbbc0220f5948047db4f78bbe1add5). If address did not have hostname it was being skipped, but kubernetes still adds dns for it, it uses the endpoint dns if hostname not specified.

UTs have been updated accordingly.  Fixed also verified against set up that reproduced the issue.

Address #167 and #155.

### DIFF
--- a/src/library.tests/WorkloadInformationProviderTests.cs
+++ b/src/library.tests/WorkloadInformationProviderTests.cs
@@ -30,19 +30,29 @@ namespace Microsoft.BridgeToKubernetes.Library.Tests
         }
 
         [Theory]
-        [InlineData(1, 100)]
-        [InlineData(5, 20)]
-        [InlineData(10, 3)]
-        [InlineData(2, 3)]
-        public async void GetReachableServicesAsync_HeadlessService(int numServices, int numAddresses)
+        [InlineData(1, 100, true)]
+        [InlineData(5, 20, false)]
+        [InlineData(10, 3, true)]
+        [InlineData(2, 3, false)]
+        public async void GetReachableServicesAsync_HeadlessService(int numServices, int numAddresses, bool isInWorkloadNamespace)
         {
-            ConfigureHeadlessService(numServices: numServices, namingFunction: (i) => $"myapp-{i}", numAddresses: numAddresses, addressHostNamingFunction: (i) => $"Host-{i}");
-            var result = await _workloadInformationProvider.GetReachableEndpointsAsync(namespaceName: "", localProcessConfig: null, includeSameNamespaceServices: true, cancellationToken: default(CancellationToken));
+            var expected = ConfigureHeadlessService(numServices: numServices, namingFunction: (i) => $"myapp-{i}", numAddresses: numAddresses, addressHostNamingFunction: (i) => $"Host-{i}", isInWorkloadNamespace);
+            var result = await _workloadInformationProvider.GetReachableEndpointsAsync(namespaceName: isInWorkloadNamespace ? "testNamespace" : "", localProcessConfig: null, includeSameNamespaceServices: true, cancellationToken: default(CancellationToken));
             // Doing numServices-1 when calculating because we are adding empty subset for one service and that will be skipped
-            Assert.Equal(numAddresses == 0 ? numServices : (numServices-1) * (numAddresses), result.Count());
+            Assert.Equal((numServices-1) * (numAddresses), result.Count());
             foreach (var endpoint in result) {
                 if (endpoint.Ports.Any()) {
                     Assert.Equal(endpoint.Ports.ElementAt(0).LocalPort, -1);
+                    bool found = false;
+                    foreach (var dns in expected) {
+                        if (string.Equals(endpoint.DnsName, dns))
+                        {
+                            found = true;
+                            expected.Remove(dns);
+                            break;
+                        }
+                    }
+                    Assert.True(found);
                 }
             }  
         }
@@ -281,8 +291,9 @@ namespace Microsoft.BridgeToKubernetes.Library.Tests
             A.CallTo(() => _autoFake.Resolve<IKubernetesClient>().ListServicesInNamespaceAsync(default, default, default)).WithAnyArguments().Returns(new V1ServiceList(serviceList));
         }
 
-        private void ConfigureHeadlessService(int numServices, Func<int, string> namingFunction, int numAddresses, Func<int, string> addressHostNamingFunction)
+        private List<string> ConfigureHeadlessService(int numServices, Func<int, string> namingFunction, int numAddresses, Func<int, string> addressHostNamingFunction, bool isInWorkloadNamespace)
         {
+            var result = new List<string>();
             var serviceList = new List<V1Service>();
             // introducing this variable so we can add an endpoint with empty subset to have crash coverage
             bool addSubset = false;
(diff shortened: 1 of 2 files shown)


src/library.tests/WorkloadInformationProviderTests.cs
```diff
@@ -33,13 +33,13 @@ public WorkloadInformationProviderTests()
...
         [InlineData(10, 3)]
-        [InlineData(1, 3)]
+        [InlineData(2, 3)]
         public async void GetReachableServicesAsync_HeadlessService(int numServices, int numAddresses)
         {
             ConfigureHeadlessService(numServices: numServices, namingFunction: (i) => $"myapp-{i}", numAddresses: numAddresses, addressHostNamingFunction: (i) => $"Host-{i}");
             var result = await _workloadInformationProvider.GetReachableEndpointsAsync(namespaceName: "", localProcessConfig: null, includeSameNamespaceServices: true, cancellationToken: default(CancellationToken));
             // Doing numServices-1 when calculating because we are adding empty subset for one service and that will be skipped
-            Assert.Equal((numServices-1) * (numAddresses), result.Count());
+            Assert.Equal(numAddresses == 0 ? numServices : (numServices-1) * (numAddresses), result.Count());
```
hsubramanianaks, line 42: I don't see use case where numAddress is zero from InlineData, is it made zero somewhere else. ? Please let me know.

src/library.tests/WorkloadInformationProviderTests.cs
```diff
@@ -33,13 +33,13 @@ public WorkloadInformationProviderTests()
         [InlineData(1, 100)]
         [InlineData(5, 20)]
         [InlineData(10, 3)]
-        [InlineData(1, 3)]
+        [InlineData(2, 3)]
         public async void GetReachableServicesAsync_HeadlessService(int numServices, int numAddresses)
         {
             ConfigureHeadlessService(numServices: numServices, namingFunction: (i) => $"myapp-{i}", numAddresses: numAddresses, addressHostNamingFunction: (i) => $"Host-{i}");
             var result = await _workloadInformationProvider.GetReachableEndpointsAsync(namespaceName: "", localProcessConfig: null, includeSameNamespaceServices: true, cancellationToken: default(CancellationToken));
             // Doing numServices-1 when calculating because we are adding empty subset for one service and that will be skipped
```
hsubramanianaks, line 41: Is this comment relevant now, because now dns name can be empty for headless services?

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -480,21 +480,33 @@ private int GetServicePortValue(V1ServicePort servicePort, V1Container container
...
-                    foreach (var address in addresses)
+                    
+                    foreach (var address in  subset.Addresses)
                     {
+                        if (address == null) {
+                            continue;
+                        }
+                        string dns = "";
+                        // If hostname is empty for the address, then dns used is that of the endpoint
+                        if (!string.IsNullOrWhiteSpace(address.Hostname))
```
hsubramanianaks, line 495: This is suggestion: we can do as a switch statement here instead of if statements as the user suggested ?

src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -480,21 +480,33 @@ private int GetServicePortValue(V1ServicePort servicePort, V1Container container
...
+                        {
+                            dns = isInWorkloadNamespace ?
+                                    $"{address.Hostname}.{endpoint.Metadata.Name}" :
+                                    $"{address.Hostname}.{endpoint.Metadata.Name}.{endpoint.Metadata.Namespace()}";
+                        }
+                        else
+                        {
+                            dns = isInWorkloadNamespace ?
```
hsubramanianaks, line 503: can you add UT for this if/else code block? ex: Assert on the List<EndpointInfo> using CompareLists or for each of this and compare the values ?

hsubramanianaks (changes requested):
Left some comments and do you think we should add a separate task/story to create todo-app with headless service so that this can be tested as well by testing team members ? Since we are seeing bugs around this often. Please let me know your thoughts. Thanks.

src/library.tests/WorkloadInformationProviderTests.cs
elenavillamil: Ah this was left over from when I was working on the UT. Will remove.

src/library/Connect/WorkloadInformationProvider.cs
elenavillamil: I decided agains switch because number of conditions is not large and readability seemed better this way. https://www.geeksforgeeks.org/switch-vs-else/ this is some reading I did on switch vs if. It seems switch is preffer if number if else nested cases is large.

src/library/Connect/WorkloadInformationProvider.cs
elenavillamil: done

elenavillamil:
> Left some comments and do you think we should add a separate task/story to create todo-app with headless service so that this can be tested as well by testing team members ? Since we are seeing bugs around this often. Please let me know your thoughts. Thanks.

Yes, I think the task should be added under E2E testing story. Once we have automated testing it will be a lot easier to add more coverage of this type of edge cases. But I dont really want to add more load on testing team atm.

src/library.tests/WorkloadInformationProviderTests.cs
```diff
@@ -30,19 +30,29 @@ public WorkloadInformationProviderTests()
...
-        public async void GetReachableServicesAsync_HeadlessService(int numServices, int numAddresses)
+        [InlineData(1, 100, true)]
+        [InlineData(5, 20, false)]
+        [InlineData(10, 3, true)]
+        [InlineData(2, 3, false)]
+        public async void GetReachableServicesAsync_HeadlessService(int numServices, int numAddresses, bool isInWorkloadNamespace)
         {
-            ConfigureHeadlessService(numServices: numServices, namingFunction: (i) => $"myapp-{i}", numAddresses: numAddresses, addressHostNamingFunction: (i) => $"Host-{i}");
-            var result = await _workloadInformationProvider.GetReachableEndpointsAsync(namespaceName: "", localProcessConfig: null, includeSameNamespaceServices: true, cancellationToken: default(CancellationToken));
+            var expected = ConfigureHeadlessService(numServices: numServices, namingFunction: (i) => $"myapp-{i}", numAddresses: numAddresses, addressHostNamingFunction: (i) => $"Host-{i}", isInWorkloadNamespace);
```
hsubramanianaks, line 39: can we change the variables expected to expectedServicesList and instead of result, finalServicesList since many of the variables are defined ```var``` it is confusing when someone reads the code. Just a nice to have suggestion. Other than than looks good, I like the idea of covering all different use cases programatically in the test using InlineData :). 

or another suggestion instead of var result, you can change it to ```IEnumerable<EndpointInfo> result ``` , same for expected variable. 

Thanks.

hsubramanianaks (approved):
LGTM except a minor comment,

src/library.tests/WorkloadInformationProviderTests.cs
elenavillamil: sure, done.

hsubramanianaks (approved):
LGTM

PR #191, opened 2023-03-08
Removes package references in favour of FrameworkReference 

### LINKED ISSUE #185: PackageReference cleanup?
After the migration to .Net 6 some of the PackageReference are no longer needed since projects that target the Microsoft.NET.Sdk.Web SDK implicitly reference the Microsoft.AspNetCore.App framework.

Is there a need to target specific versions of the Microsoft packages or should these references be removed in favour of the shared framework versions?

I think using the shared framework is better from a security perspective since it will pick the version from the installed SDK on the build server.

I can make a PR for this if you want.

Eneuman:
This PR removes packagereferences in favour of FrameworkReference (shared libarary).

This means that B2K will no longer use a specific version of a .net package but instead use the version of the package that is installed by the runtime on the client computer or in the docker image. This has the benefit that the contents will automatically update when the runtime is updated.

An "assert" in a test was also changed so the reference to the fluent validation NuGet package could be removed.

The following packages was removed because they are already included in the FrameworkReference meta package:
- Microsoft.AspNetCore.Http
- Microsoft.AspNetCore.Mvc.Core
- Microsoft.Extensions.Hosting.Abstractions

The following package were removed because they were already included as a dependency:
- Microsoft.IdentityModel.Tokens
- Microsoft.Win32.Registry
- System.IO.FileSystem.AccessControl
- System.Security.Principal
- System.Text.Encodings.Web
- System.Text.Json

This PR also changes the Sdk for the DSC project from Microsoft.NET.Sdk.Web to Microsoft.NET.Sdk because creating nuget packages from web projects are not supported.

fixes #185


src/routingmanager/routingmanager.csproj
```diff
@@ -25,9 +25,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Autofac" Version="6.4.0" />
```
hsubramanianaks, line 28: Thanks @Eneuman  for this PR. are you sure removing Autofac package reference doesn't impact DI ? was this tested locally ?

Eneuman: @hsubramanianaks 
Autofac is implicit reference from the Autofac.Extensions.DependencyInjection package so it's available to the code :)

hsubramanianaks: Thanks for confirming on this.

src/devhostagent/devhostAgent.csproj
```diff
@@ -34,8 +34,6 @@
 
   <ItemGroup>
     <PackageReference Include="Autofac.Extensions.DependencyInjection" Version="8.0.0" />
-    <PackageReference Include="Microsoft.AspNetCore.SignalR.Protocols.MessagePack" Version="6.0.8" />
```
hsubramanianaks, line 37: Any idea why signalR packages were removed ? Was this intentional ? Please let us know.Thanks.

Eneuman: Yes that was intentionally. SignalR packages was already included in the dependent common project.

I couldn't find any documentation describing the desired way of handling references to NuGet packages in project files so I went for the "first project in dependency graph that uses a package have the dependency".

I'm also fine with "any project using a dependency have the dependency" approach, just let me know how you want it and I'll fix it :)

elenavillamil:
@Eneuman, my only concern with this PR is if we have to worry about braking changes between versions since now the versions for some libraries used during testing may be different from what the user is running with.

Eneuman:
@elenavillamil yeah technically that can happen, but I think the chances of that happening is very low since it's the default behaviour of any application built using the default templates in VS 2022 :)

The benefits from using the shared framework are great in regards to performance improvements, from precompiled assets, and the ability to apply in-place security updates through windows update.

Eneuman:
There are also a lot if checks in place:

Using the Microsoft.AspNetCore.App metapackage provides version restrictions that protect your app:
* If a package is included that has a transitive (not direct) dependency on a package in Microsoft.AspNetCore.App, and those version numbers differ, NuGet will generate an error.
* Other packages added to your app cannot change the version of packages included in Microsoft.AspNetCore.App.
Version consistency ensures a reliable experience. * Microsoft.AspNetCore.App was designed to prevent untested version combinations of related bits being used together in the same app.

src/common/common.csproj
```diff
@@ -46,31 +46,24 @@
   </ItemGroup>
 
   <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
hsubramanianaks, line 49: @Eneuman Bridge uses both self contained for VS extensions and framework dependent executables for VS code extensions. does this change affect self contained builds ? Did you try running ```dotnet publish -r RID --self contained true``` ?

hsubramanianaks: @Eneuman I tried myself and compared your changes with the one in the main branch without shared framework. seems like some of the packages are missing ex: AspNetCore.JsonPatch refer below for a screenshots. 

Microsoft.Extensions.DependencyModel.dll
image

Microsoft.Dotnet.PlatformAbstractions.dll
image

Microsoft.CodeAnalysis.dll
image

Microsoft.AspNetCore.JsonPatch.dll
image

also System.Security.Permissions.dll is also missing. Any idea on these ?? 

Note: also using shared framework had both advantages and disadvantages,

Eneuman: @hsubramanianaks yes, jsonpatch was removed in a earlier PR (when Newtonsoft was replaced with STJ) and the other packages are removed by the publish process because it can not find any references to them so I think it's all good. I didn't know B2K used self contained builds so I'm gonna do some extra testing around that just to make sure. 

What parts are running as self-contained?

hsubramanianaks: @Eneuman Thanks, I will send this to the testing team and see how it goes. Yes PublishReadyToRun=true seems like good option. I can do another PR on the pipelines to include them. Thanks for the suggestion. :)

hsubramanianaks:
Closing PR to kick in chatGTP PR review.

hsubramanianaks (changes requested):
@Eneuman Build is failing due to dotnet pack giving below error , can you try running this in your local to see if it works ? 

```dotnet pack src/dsc/dsc.csproj /p:NuspecFile=src/dsc/dsc.nuspec```

```"D:\a\_work\1\s\src\dsc\dsc.csproj" (pack target) (1) ->
       "D:\a\_work\1\s\src\dsc\dsc.csproj" (_GetTfmSpecificContentForPackage target) (1:3) ->
       (LoadStaticWebAssetsBuildManifest target) -> 
         D:\a\_work\_temp\dotnet\sdk\6.0.407\Sdks\Microsoft.NET.Sdk.Razor\targets\Microsoft.NET.Sdk.Razor.StaticWebAssets.targets(521,5): error : Manifest file at 'obj\Debug\net6.0\staticwebassets.build.json' not found. [D:\a\_work\1\s\src\dsc\dsc.csproj]
```

Eneuman:
@hsubramanianaks this is strange. Can you try and run `dotnet pack src/dsc/dsc.csproj /p:NuspecFile=dsc.nuspec` ?
That works for me, but notice that I excluded the path from the Nuspec argument.

Eneuman:
@hsubramanianaks What version of SDK did the build?

Eneuman:
@hsubramanianaks I think that some versions of .Net SDK cannot pack project targeting the Microsoft.NET.Sdk.Web Sdk. I know they have fixed this in newer versions of the SDK but there are some ways to get around this. Can you check the build logs and see what project is breaking the build atm ?

hsubramanianaks:
> @hsubramanianaks Some versions of .Net SDK cannot pack project targeting the Microsoft.NET.Sdk.Web Sdk. I know they have fixed this in newer versions of the SDK but there are some ways to get around this. Can you check the build logs and see what project is breaking the build atm ?

It is the main ```dsc.csproj``` which is used by nuget for vs extension to use it as nuget package. I checked the logs it uses .NET sdk 6.0.407. I can provide the [link](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_apis/build/builds/7584351/logs/37) you might not have access to it. Also it is all in the PR build Yaml [here](https://github.com/Azure/Bridge-To-Kubernetes/blob/main/build-pipelines-PR.yml#L23) which shows SDK being used. I will do dry run in my local to see if it helps. Thank you.

image

Eneuman:
@hsubramanianaks I tested it with 6.0.307 just now and that passed. 6.0.400 had a lot of problems regarding this error and MAUI, may be that's related.
I'm gonna download 6.0.400 and see if I can replicate the problem.
Could you post the part of the log showing the error please?

Eneuman:
@hsubramanianaks I did some more research and this is what I found out.

It seems like packing of a web application is not supported.

That means that we probably should change the sdk DSC uses from Microsoft.NET.Sdk.Web to Microsoft.NET.Sdk and explicit reference the needed packages in that project.

hsubramanianaks:
> @hsubramanianaks I did some more research and this is what I found out.
> 
> It seems like packing of a web application is not supported.
> 
> That means that we probably should change the sdk DSC uses from Microsoft.NET.Sdk.Web to Microsoft.NET.Sdk and explicit reference the needed packages in that project. 

@Eneuman thank you for your research. I think if you change it to non web sdk some part in dsc might not work (needs confirmation). For example dsc spins up a service with control port number which is used by both extensions to disconnect dsc and restore to original state. If you are changing it can you test in your local with todo app and test cluster or with your own applications/cluster to make sure that works ? Please let me know for further details.

Eneuman:
@hsubramanianaks I will do some more testing this weekend to make sure it works!

hsubramanianaks:
> @hsubramanianaks I will do some more testing this weekend to make sure it works!

I will do testing as well. Seems like it's Good Friday enjoy your holiday!!

hsubramanianaks:
> @hsubramanianaks I will do some more testing this weekend to make sure it works!

@Eneuman I tested this, it works. I will send it to testing team for full E2E testing. Thank you.

Eneuman:
@hsubramanianaks I did some live testing (for about 4 hours)) as well (used this branch when developing our product) and I didn't run into any problems :)

hsubramanianaks (approved):

PR #193, opened 2023-03-08
Cleanup PR build pipelines

hsubramanianaks:
This PR is to clean up PR build pipelines and make them simple. 

image

### DIFF
--- a/build-pipelines-PR.yml
+++ b/build-pipelines-PR.yml
@@ -40,7 +40,7 @@ resources:
     ref: refs/heads/main
 jobs:
 - job: Phase_1
-  displayName: Run PR build for self contained
+  displayName: Run self-contained build (binaries V1, where the binaries are all bundled together)
   timeoutInMinutes: 120
   cancelTimeoutInMinutes: 1
   pool:
@@ -93,7 +93,7 @@ jobs:
     enabled: True
     inputs:
       solution: src/client.sln
-      msbuildArgs: -p:ConsumeEndpointManager=false
+      msbuildArgs: -p:Localize=true -p:ConsumeEndpointManager=false
       configuration: ${{ parameters.BuildConfiguration }}
   - task: VSTest@2
     displayName: Run Unit Tests
@@ -162,7 +162,7 @@ jobs:
       buildProperties: 'NuspecFile=bin\${{ parameters.BuildConfiguration }}\net6.0\win-x64\publish\dsc.nuspec'
 
 - job: Phase_2
-  displayName: Run PR build for non self contained
+  displayName: Run NON self-contained build (binaries V2, where the binaries are downloaded in parallel)
   timeoutInMinutes: 120
   cancelTimeoutInMinutes: 1
   pool:
--- a/build-pipelines-PR.yml
+++ b/build-pipelines-PR.yml
@@ -40,7 +40,7 @@ resources:
     ref: refs/heads/main
 jobs:
 - job: Phase_1
-  displayName: Run PR build
+  displayName: Run PR build for self contained
   timeoutInMinutes: 120
   cancelTimeoutInMinutes: 1
   pool:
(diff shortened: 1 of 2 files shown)


build-pipelines-PR.yml
```diff
@@ -103,7 +69,6 @@ jobs:
       restoreArguments: -r win-x64
       selectOrConfig: config
       nugetConfigPath: src/nuget.config
-      externalEndpoints: dba0ba50-2d4e-4f12-9f5a-42a638da803b
```
elenavillamil, line 106: is this removal intentional, if so why is this not needed now?

build-pipelines-PR.yml
```diff
@@ -129,7 +93,7 @@ jobs:
     enabled: True
     inputs:
       solution: src/client.sln
-      msbuildArgs: -p:Localize=true -p:ConsumeEndpointManager=false
```
elenavillamil, line 132: why remove this part? This taks is the one failing when PR is sent from fork, is this the reason why?

build-pipelines-PR.yml
```diff
@@ -225,224 +151,6 @@ jobs:
...
-
-
-        New-Item -Path './src/dsc/bin/${{ parameters.BuildConfiguration }}/net6.0/linux-x64/publish/kubectl' -ItemType Directory
-
-        New-Item -Path './src/dsc/bin/${{ parameters.BuildConfiguration }}/net6.0/linux-x64/publish/kubectl/linux' -ItemType Directory
-
-        curl https://dl.k8s.io/release/v1.21.2/bin/linux/amd64/kubectl  -OutFile ./src/dsc/bin/${{ parameters.BuildConfiguration }}/net6.0/linux-x64/publish/kubectl/linux/kubectl
-
-  - task: CopyFiles@2
-    displayName: Collect files for .zip (Windows)
```
elenavillamil, line 256: Simpler is great, but for some of these tasks I wonder if it is worth keeping parity with release pipeline. So PR build and release pipeline stay as close as possible aside from tasks that cant be ran from fork. I am open either way, just thinking.

hsubramanianaks: This will reduce the time PR builds take and it is already validating if it restores, builds and UT's. I think these copy files, zip tasks wouldn't change much. if you think if we want to keep parity then we can't save much on the build time. Just a suggestion on pros/cons of doing it simple.

build-pipelines-PR.yml
hsubramanianaks: so on the other parameter ConsumeEndpointManger is false because we are building/publishing endpoint manager as separate step in the pipelines so that main dsc project doesn't do it again. By default value is true. 

image

hsubramanianaks: The failures for the forked repos are related to the external endpoints being present in the restore endpoint manager step above https://github.com/Azure/Bridge-To-Kubernetes/pull/193/files#r1131937016. That's why this step is failing, please see below screenshot. 

image

image

build-pipelines-PR.yml
hsubramanianaks: here is the reference to the document 
image

build-pipelines-PR.yml
elenavillamil: sounds good. let go with removing them.

build-pipelines-PR.yml
```diff
@@ -452,125 +160,54 @@ jobs:
...
+  displayName: Run PR build for non self contained
   timeoutInMinutes: 120
   cancelTimeoutInMinutes: 1
   pool:
     name: 'VSEngSS-MicroBuild2022-1ES'
     demands:
     - msbuild
     - visualstudio
-    - vstest
+    - DotNetFramework
```
elenavillamil, line 173: intentional? Curious why.

build-pipelines-PR.yml
```diff
@@ -51,7 +40,7 @@ resources:
     ref: refs/heads/main
 jobs:
 - job: Phase_1
-  displayName: Run self-contained build (binaries V1, where the binaries are all bundled together)
+  displayName: Run PR build for self contained
```
elenavillamil, line 43: nit and maybe just my personal reference but i like having the link between self contained and binaries V1. If possible I liked previous display name better. okay to update display name if we add a comment as well.

elenavillamil (approved):
Looks good. Left two minor comments.

build-pipelines-PR.yml
hsubramanianaks: I kept it similar to self-contained one's if you see above, it has all these three demands.

build-pipelines-PR.yml
hsubramanianaks: I changed it back, since it was long, and it was cutting off in the ADO UI that's why made it short.

build-pipelines-PR.yml
elenavillamil: ok we may want to update release pipeline too then?

elenavillamil (approved):
LGMT

PR #196, opened 2023-03-09
Updating database api project to dotnet 6

hsubramanianaks:
This PR is to fix the deployment issue for database api in samples/todo/database-api due to dotnet version mismatch between csproj and docker file. 

image

### DIFF
--- a/samples/todo-app/database-api/databaseApi.csproj
+++ b/samples/todo-app/database-api/databaseApi.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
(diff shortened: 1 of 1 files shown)


samples/todo-app/database-api/databaseApi.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
```
hsubramanianaks, line 4: I tested this with VS 2019 and VS 2022 B2k extensions , both work fine. Only catch with VS 2019 is if they are using this application to test they have to select ```use preview of .NET SDK's. (requires restart)``` in Options in VS2019 would make it work. 

image

elenavillamil: Seems preview version fixes the bug we had, that we ran into in December :) very nice!

elenavillamil (approved):

Issue #199, opened 2023-03-13
Restore job fails at times due to  spec.template.spec.containers[0].env[0].name: Required value


hsubramanianaks:
**Describe the bug**
Restoration job failed due to 
 spec.template.spec.containers[0].env[0].name: Required value 

**Mention the platform you are using**
VSCode, AKS cluster

**To Reproduce**
1. Create a K8s cluster
2. Debug a service/pod which fails the deployment
3. the pod doesn't go to the original state and fails to restore. 

**Expected behavior**
The pod should go to the original state in case of exception.
**Screenshots**
Format":"Restore: Failed to restore Deployment 'database-api' to original state with image 'azdspublic/todo-app-database-api'. Unprocessable Entity: Deployment.apps \"database-api\" is invalid: spec.template.spec.containers[0].env[0].name: Required value","Args":[],"Message":"Restore: Failed to restore Deployment 'database-api' to original state with image 'azdspublic/todo-app-database-api'. Unprocessable Entity: Deployment.apps \"database-api\" is invalid: spec.template.spec.containers[0].env[0].name: Required value

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Chrome
 - Version Production

**Additional context**

hsubramanianaks:
Closing this as this is not happening anymore.

PR #203, opened 2023-03-14
Restore job failure fix

hsubramanianaks:
This PR is to fix the following issue, this change improvises to make sure even in case of devhost agent connected sessions api doesn't return any proper response, in that case lastPingWithSessions will be null. in that case, we will restore back to original state. 

https://github.com/Azure/Bridge-To-Kubernetes/issues/199


deployment/settings/services/imagetag.setting
```diff
@@ -1,4 +1,4 @@
 MINDARO_DEVHOSTAGENT_TAG=1.1.0
-MINDARO_DEVHOSTAGENT_RESTORATIONJOB_TAG=1.1.0
+MINDARO_DEVHOSTAGENT_RESTORATIONJOB_TAG=1.2.0
```
hsubramanianaks, line 2: Updating this based off this read me, so that newer version of restoration job will be pulled. - https://github.com/Azure/Bridge-To-Kubernetes/tree/main/src/devhostagent#versioning

src/devhostagent.restorationjob/RestorationJobApp.cs
```diff
@@ -142,10 +142,17 @@ private async Task<ExitCode> ExecuteInnerAsync(CancellationToken cancellationTok
                             perfLogger.SetProperty(HasConnectedClients, false);
 
                             var disconnectedTimeSpan = DateTimeOffset.Now - lastPingWithSessions;
-                            if (disconnectedTimeSpan != null && disconnectedTimeSpan.Value > _restorationJobEnvironmentVariables.RestoreTimeout)
+                            if ((disconnectedTimeSpan != null && disconnectedTimeSpan.Value > _restorationJobEnvironmentVariables.RestoreTimeout) || lastPingWithSessions == null)
```
elenavillamil, line 145: I think we should add a count check here for safety, just in case when bridge is starting this condition may ever happen first time around. If lastPingWithSessions has been null with no connected sessions 3 times then restore. It solves issue same way just in third iteration of this loop.

src/devhostagent.restorationjob/RestorationJobApp.cs
```diff
@@ -142,10 +144,10 @@ private async Task<ExitCode> ExecuteInnerAsync(CancellationToken cancellationTok
                             perfLogger.SetProperty(HasConnectedClients, false);
 
                             var disconnectedTimeSpan = DateTimeOffset.Now - lastPingWithSessions;
-                            if (disconnectedTimeSpan != null && disconnectedTimeSpan.Value > _restorationJobEnvironmentVariables.RestoreTimeout)
+                            if ((disconnectedTimeSpan != null && disconnectedTimeSpan.Value > _restorationJobEnvironmentVariables.RestoreTimeout) || (lastPingWithSessions == null && count > 2))
```
elenavillamil, line 147: nit: count is generic name, it maybe better timesLastPingWithSessionsIsNull? or something like that.

src/devhostagent.restorationjob/RestorationJobApp.cs
```diff
@@ -142,10 +144,10 @@ private async Task<ExitCode> ExecuteInnerAsync(CancellationToken cancellationTok
                             perfLogger.SetProperty(HasConnectedClients, false);
 
                             var disconnectedTimeSpan = DateTimeOffset.Now - lastPingWithSessions;
-                            if (disconnectedTimeSpan != null && disconnectedTimeSpan.Value > _restorationJobEnvironmentVariables.RestoreTimeout)
+                            if ((disconnectedTimeSpan != null && disconnectedTimeSpan.Value > _restorationJobEnvironmentVariables.RestoreTimeout) || (lastPingWithSessions == null && count > 2))
                             {
                                 // Restore workload
-                                _log.Info($"Agent has no connected sessions for {disconnectedTimeSpan.Value:g}. Restoring...");
```
elenavillamil, line 148: removal or :g intentional?

src/devhostagent.restorationjob/RestorationJobApp.cs
hsubramanianaks: @elenavillamil Thanks for the suggestion. Instead of fixed count, I thought we will rely on existing restore timeout and calculate the time until lastPingWithSessions is null and if that exceeds restore timeout then it would restore back to original state. Updated UT's accordingly. Please let me know.

src/devhostagent.restorationjob/RestorationJobApp.cs
hsubramanianaks: replaced count with ```timeUntilLastPingIsNull```

src/devhostagent.restorationjob/RestorationJobApp.cs
hsubramanianaks: Yes this was causing not valid time error, but after refactoring the code, the purpose of changing this statement is not needed.

src/devhostagent.restorationjob/RestorationJobApp.cs
```diff
@@ -91,6 +91,7 @@ private async Task<ExitCode> ExecuteInnerAsync(CancellationToken cancellationTok
                 await Task.Delay(_restorationJobEnvironmentVariables.PingInterval, cancellationToken);
                 int numFailedPings = 0;
                 DateTimeOffset? lastPingWithSessions = null;
+                DateTimeOffset? timeUntilLastPingIsNull = null;
```
elenavillamil, line 94: nit: timeSinceLastPingIsNull ?

src/client.sln
```diff
@@ -26,6 +26,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "endpointmanagerlauncher", "
 EndProject
```
elenavillamil, line 26: nit: changes on this file needed?

elenavillamil (approved):
minor comments.

src/client.sln
hsubramanianaks: Yes needed to run UT's via build systems, I am not sure why dsc.tests, routingmanager.tests, restoration.tests were excluded previously but I am going to add them and see how it goes.

elenavillamil (approved):
Looks good to me :)

src/devhostagent.restorationjob/RestorationJobApp.cs
hsubramanianaks: updated.

Issue #143, opened 2023-01-10
Can't connect cluster service to debug in the connected  dev machine


GKilm:
**Describe the bug**

The program being debugged by B2K can't connect other services in the cluster.

Here are the environment variable in the debugged pod and the dev computer.

the debugged pod
``` shell
export DEFAULT_ANSIBLESERVER_PORT='tcp://10.98.132.103:30890'
export DEFAULT_ANSIBLESERVER_PORT_30890_TCP='tcp://10.98.132.103:30890'
export DEFAULT_ANSIBLESERVER_PORT_30890_TCP_ADDR='10.98.132.103'
export DEFAULT_ANSIBLESERVER_PORT_30890_TCP_PORT='30890'
export DEFAULT_ANSIBLESERVER_PORT_30890_TCP_PROTO='tcp'
export DEFAULT_ANSIBLESERVER_SERVICE_HOST='10.98.132.103'
export DEFAULT_ANSIBLESERVER_SERVICE_PORT='30890'
export DEFAULT_ANSIBLESERVER_SERVICE_PORT_API='30890'
export DEFAULT_APIGATEWAY_PORT='tcp://10.106.234.148:30300'
export DEFAULT_APIGATEWAY_PORT_30300_TCP='tcp://10.106.234.148:30300'
export DEFAULT_APIGATEWAY_PORT_30300_TCP_ADDR='10.106.234.148'
export DEFAULT_APIGATEWAY_PORT_30300_TCP_PORT='30300'
export DEFAULT_APIGATEWAY_PORT_30300_TCP_PROTO='tcp'
export DEFAULT_APIGATEWAY_PORT_30443_TCP='tcp://10.106.234.148:30443'
export DEFAULT_APIGATEWAY_PORT_30443_TCP_ADDR='10.106.234.148'
export DEFAULT_APIGATEWAY_PORT_30443_TCP_PORT='30443'
export DEFAULT_APIGATEWAY_PORT_30443_TCP_PROTO='tcp'
export DEFAULT_APIGATEWAY_SERVICE_HOST='10.106.234.148'
export DEFAULT_APIGATEWAY_SERVICE_PORT='30300'
export DEFAULT_APIGATEWAY_SERVICE_PORT_API='30300'
export DEFAULT_APIGATEWAY_SERVICE_PORT_WS='30443'
export DEFAULT_CLOUDEVENT_PORT='tcp://10.103.129.124:30892'
[...]

PR #206, opened 2023-03-16
Support nodeport service

elenavillamil:
This PR targets #143. Today bridge sets up port forwarding for cluster ip and LoadBalancer, but skips NodePort. After this PR bridge will also set up forwarding for services of type NodePort. I have added UTs and manually tested it by making stats-cache service of type NodePort in todo-sample app.
Screenshot 2023-03-16 at 10 13 57 AM


hsubramanianaks (approved):
looks good to me. Thanks.

PR #208, opened 2023-03-17
Updating logging property and update string manipulation

elenavillamil:
We need to update failure reason property name since existing one was suppressed. Also adding UTs and fix function per UTs.


src/common.tests/Utilities/StringManipulationTests.cs
```diff
@@ -0,0 +1,25 @@
...
+
+using Microsoft.BridgeToKubernetes.Common.Utilities;
+using Xunit;
+
+namespace Microsoft.BridgeToKubernetes.Common.Tests.Utilities
+{
+    public class StringManipulationTests
+    {
+        [Theory]
+        [InlineData("BEGIN PRIVATE KEY\nuirethweuifhewiufhweiuofhweo\njweiorjweoifnewouhfousdgherufhsdiuhEND PRIVATE KEY\n", "KEY_WAS_REMOVED\n")]
```
hsubramanianaks, line 14: can you add one more InlineData for ```BEGIN PRIVATE KEYb2kkkkkkkkkkkkkkEND PRIVATE KEY``` without ```\n``` break lines ? just to make sure different use cases work.

hsubramanianaks (approved):
Looks good to me.

PR #209, opened 2023-03-19
Refactoring Json serialization

### LINKED ISSUE #204: Refactor static Json Serialization functionallity to enable logging and easier unit testing
Today serialization to Json is done in two different ways, using the `Serialize` function in `logging/extension.cs` (common project) and using the functions in `JsonHelper.cs` (also common project).

Since both theese classes are static, no logging is possible and they makes it harder to create unit tests.

I suggest:

- moving the serialization functionallity into the class JsonHelpers
- add a interface to the class
- make the class non static
- add logging to any exception that occurs
- update the code to use the class through DI.

Pros:
* Exceptions that occurs will be logged to both to AI and local logs (i suggest logging them as warnings since they are swallowed today)
* It will be easier to create tests since you can now mock the serialization.
* Since all serialization exceptions are logged it's easier to debug some problems and edge cases.

Cons:
* Some existing code patterns creates TaskCanceled exceptions today and theese will be logged as warnings

I can make a PR for this it you want.

Eneuman:
This PR aims to make it easier to understand what serialization function should be used and where.

**Serialization used for logging**
If this serialization fails it should never crash the program.
Two methods handel this:
```
string SerializeForLoggingPurpose(object input)
string SerializeForLoggingPurposeIndented(object input)
```

**Serialization used for running the program**
If this serialization fails it should be logged by the function calling the serialization. 
Four methods handel this:
```
T DeserializeObject<T>(string v)
T DeserializeObjectCaseInsensitive<T>(string v)
string SerializeObject(object obj)
string SerializeObjectIndented(object obj)
```

**Other fixes**
- Lowered the memory allocations by making sure JsonSerilizationOptions is only created once


hsubramanianaks:
Closing PR to kick in chatgpt review.

src/common/Models/Kubernetes/Pod.cs
```diff
@@ -1,95 +0,0 @@
...
-using System.Linq;
-using Microsoft.BridgeToKubernetes.Common.Json;
-using Microsoft.BridgeToKubernetes.Common.Logging;
-
-namespace Microsoft.BridgeToKubernetes.Common.Models.Kubernetes
-{
-    /// <summary>
-    /// Kubernetes Pod
-    /// </summary>
-    public class Pod
```
hsubramanianaks, line 17: @Eneuman This file was deleted intentionally ? I see it doesn't have any usage but maybe we can do another PR to use remove such files ? a PR to remove unused files. Just a thought, please let me know thank you.

src/common.tests/ConsoleOutputTests.cs
```diff
@@ -56,7 +56,7 @@ public void SingleItem()
             const string ExpectedTestRow = "name1          Name2  name3";
             const string ExpectedTestItemJson = @"[
               {
-                ""name1"": ""name1"",
```
hsubramanianaks, line 59: @Eneuman the format is changed due to STJ changes ? Trying to understand why this is changed.

Eneuman: @hsubramanianaks Nice catch, STJ uses PascalCase formating as default and I had missed adding PropertyNamingPolicy when console logging json. This has now been added.

Originaly I change the test because there was no `JsonPropertyName` attribute on the Name1 property in `TestOutputData` so I tought the test was wrong.

Theese changes only affect logging json to the console window.

src/common/Extensions/HttpExtensions.cs
```diff
@@ -7,7 +7,6 @@
 using Microsoft.BridgeToKubernetes.Common;
 using Microsoft.BridgeToKubernetes.Common.Json;
 using System.Linq;
-using System.Threading.Tasks;
 
 namespace System.Net.Http
```
hsubramanianaks, line 11: @Eneuman I don't see this class file is used anywhere can we remove this, or I saw another PR coming for unused code is it covered in that ?

src/common.tests/JsonHelperTests.cs
```diff
@@ -61,24 +55,6 @@ public void SerializeObjectWithReferenceLoopDefaultSettings()
...
-            Person p2 = new Person()
-            {
-                Name = "Samwise",
-                BestFriend = p1
-            };
-
-            p1.BestFriend = p2;
-            Assert.Throws<JsonException>(() => JsonHelpers.SerializeObject(p2, serializerSettings));
-        }
-
```
hsubramanianaks, line 81: Can we add new UT's for the new methods in JsonHelper class please ? ex: ```SerializeForLoggingPurposeIndented, SerializeForLoggingPurpose``` ?

src/common/Json/JsonHelpers.cs
```diff
@@ -10,100 +10,141 @@
...
             }
 
-            return jsonSerializerOptions;
+            try
+            {
+                return JsonSerializer.Serialize(input, settings);
+            }
+            catch (Exception)
+            {
+                return "Serialization Error";
```
hsubramanianaks, line 124: @Eneuman I have seen this happening quite a few times in the logs, is there any way we can avoid this or improvise this with ExceptionConverters, right now exception converters does only remove ```TargetSite.Module```. Suggestions or thoughts please ?

src/common.tests/Logging/ExtensionsTests.cs
```diff
@@ -23,30 +23,6 @@ public class ExtensionsTests
                 { "Authorization", new List<string>() { "authorization" } },
             };
 
-        [Fact]
-        public void SerializeAggregateException()
```
hsubramanianaks, line 27: @Eneuman since this was removed from ```Extensions.cs``` and added to JsonHelper, can you add them to JsonHelperTests please ?

hsubramanianaks (changes requested):
@Eneuman overall looks good, I would suggest adding more UT's for Json serialize and deserialize for logging purpose both indented and non indented to cover different use case ex: like [issue](https://github.com/Azure/Bridge-To-Kubernetes/issues/246#issuecomment-1520877879). seems like Live tests passed so we don't have to wait for testing team we can merge to main and they would test them in main. :)

src/common.tests/JsonHelperTests.cs
```diff
@@ -61,24 +55,6 @@ public void SerializeObjectWithReferenceLoopDefaultSettings()
             Assert.Equal(expected, JsonHelpers.SerializeObject(p2));
         }
 
-        [Fact]
```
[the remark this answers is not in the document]
Eneuman: Yes, this was removed because no code used ReferenceHandler == null

src/common.tests/JsonHelperTests.cs
Eneuman: Added new tests for more senarios :)

src/common.tests/Logging/ExtensionsTests.cs
Eneuman: @hsubramanianaks Fixed

src/common/Json/JsonHelpers.cs
Eneuman: @hsubramanianaks Yeah, I have seen this aswell. I think the main problem why theese are written to the log is because of how CancelationExceptions are handled. Today they are handled as regular exceptions and hence written to the log. The problem is that CancelationException cannot be serialized and thats why "Serialization Error" occurs.
After this PR, "Serialization Error" can only occure when writing objects to the logsfile, so they can safely be ignored until CancelationException handling is refactored.

I could change the output to "Serialization Error of type: {input.GetType()} when trying to write to logfile." to make it more precise?

hsubramanianaks: @Eneuman Yes that would help a lot. Apologies for late reply.

src/common/Extensions/HttpExtensions.cs
hsubramanianaks: @Eneuman If they are use then it is fine.

src/common/Json/JsonHelpers.cs
Eneuman: @hsubramanianaks Fixed!

hsubramanianaks (approved):
looks good, testing passed.

PR #214, opened 2023-03-21
New feature: Disable Lifecycle Hooks in remote agent

### LINKED ISSUE #87: lifecycle hooks cause issues
**Describe the bug**
The container that is spun up (isolated) copies across the lifecycle hooks from the container of context. In this instance, hooks dependent on paths for example will fail.

**Mention the platform you are using**
Not platform specific, but verified against Mac, Windows, Debian all the same

**To Reproduce**
1. Create a container with lifecycle hooks, i.e. `postStart`
2. Have that lifecycle hook depend on a path built into your container, i.e. `WORKDIR /foo/bar`
3. Run the bridge and get `CrashLoopBackOff` on the container.

**Expected behavior**
If the container does not respect the same working directories as the target container, then copying lifecycle hooks across is less than ideal.

**Screenshots**
```sh
$ kubectl -n my-ns get pod isolated-foo-78d99c4c56-8zjsh -o yaml
...
    image: bridgetokubernetes.azurecr.io/lpkremoteagent:0.1.8
    imagePullPolicy: IfNotPresent
    lifecycle:
      postStart:
        exec:
          command:
          - /bin/sh
          - -c
          - /foo/bar/bin/run migrations:migrate && /foo/bar/bin/run cache:clear
      preStop:
        exec:
          command:
          - sleep
          - "90"
 ...
 ```

**Additional context**
Philosophical differences aside re: how `postStart` is being used here, the fact that we are potentially dependent on non-existing paths is problematic.

### PR DESCRIPTION
**Pull request by Eneuman**
This PR disables lifecycle hooks from beeing cloned to the remote agent pod when the remote agent is deployed.
Existing lifecycle hooks are restored when EndpointManager disconnects.

You can enable cloning of lifecycle hooks by adding the following line to the KubernetesLocalProcessConfig.yaml file.

```
enableFeatures:
  - LifecycleHooks  

```

Fixes #87

### DIFF
--- a/src/library/Connect/Environment/EnableFeature.cs
+++ b/src/library/Connect/Environment/EnableFeature.cs
@@ -8,6 +8,7 @@ namespace Microsoft.BridgeToKubernetes.Library.Connect.Environment
     internal enum EnableFeature
     {
         ManagedIdentity,
-        Probes
+        Probes,
+        LifecycleHooks
     }
 }
--- a/src/library/Connect/Environment/ILocalProcessConfig.cs
+++ b/src/library/Connect/Environment/ILocalProcessConfig.cs
@@ -55,6 +55,11 @@ namespace Microsoft.BridgeToKubernetes.Library.Connect.Environment
         bool IsProbesEnabled { get; }
 
         /// <summary>
+        /// Is Container Lifecycle Hooks one of the enabled features
+        /// </summary>
+        bool IsLifecycleHooksEnabled { get; }
+
+        /// <summary>
         /// Returns a collection with the environment variables generated from the local env file
         /// </summary>
         /// <returns></returns>
--- a/src/library/Connect/Environment/LocalProcessConfig.cs
+++ b/src/library/Connect/Environment/LocalProcessConfig.cs
@@ -148,6 +148,11 @@ namespace Microsoft.BridgeToKubernetes.Library.Connect.Environment
                         {
                             this.IsProbesEnabled = true;
                         }
+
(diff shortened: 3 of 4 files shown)


hsubramanianaks (approved):
looks good. @Eneuman Thanks for this change. I will send it to testing team next week. I know there are no UT's existing , is it possible to add one ? Thanks.

Eneuman:
@hsubramanianaks Unfortunately it's not easy to add UTs for this unless we first refactor the code quite a bit. I had the same problem when added the probes feature. 

I can refactor the code and add UTs in another PR if you want?

hsubramanianaks:
> @hsubramanianaks Unfortunately it's not easy to add UTs for this unless we first refactor the code quite a bit. I had the same problem when added the probes feature.
> 
> I can refactor the code and add UTs in another PR if you want?

we added UT's for the same file in this [PR](https://github.com/Azure/Bridge-To-Kubernetes/pull/178/files) , may be you can see if this helps? What kind of refactoring is needed could you please explain more on it ? Thank you :)

Eneuman:
> we added UT's for the same file in this [PR](https://github.com/Azure/Bridge-To-Kubernetes/pull/178/files) , may be you can see if this helps? What kind of refactoring is needed could you please explain more on it ? Thank you :)

Ahh, Nice! 
Thats a very clever way of injecting the asserts :)
I'll wait until that PR is merged then i'll use the same method to add my tests to that file :)

hsubramanianaks (approved):
looks good to me.

PR #227, opened 2023-03-30
226 - add ns to env vars when workload ns != service ns

garyburgmann:
This PR is to add namespace to the env variable for the services which are in not in workload's namespace and defined via KubernetesLocalProcessConfig.yaml.


src/library/Connect/LocalEnvironmentManager.cs
```diff
@@ -436,12 +436,26 @@ private void _ReportProgress(EventLevel eventLevel, string message, params objec
 
                 // because we are using dns name instead of service we have to retrieve it by splitting when needed
                 // If this ever cause issues we should consider larger refactor where we add serviceName member variable to EndpointInfo class.
-                var serviceName = endpoint.DnsName
+                var dnsNameArray = endpoint.DnsName
                     .ToUpperInvariant()
-                    .Split(".")
```
garyburgmann, line 441: this was redundant regardless of my changes

src/library/Connect/LocalEnvironmentManager.cs
```diff
@@ -436,12 +436,26 @@ private void _ReportProgress(EventLevel eventLevel, string message, params objec
...
-                var serviceName = endpoint.DnsName
+                var dnsNameArray = endpoint.DnsName
                     .ToUpperInvariant()
-                    .Split(".")
+                    .Split(".");
+
+                var serviceName = dnsNameArray
                     .First()
-                    .Replace("-", "_")
-                    .Replace(".", "_");
```
garyburgmann, line 444: this was redundant regardless of my changes

src/library/Connect/LocalEnvironmentManager.cs
```diff
@@ -436,12 +436,26 @@ private void _ReportProgress(EventLevel eventLevel, string message, params objec
...
+
+                // when !endpoint.IsInWorkloadNamespace
+                var serviceNs = dnsNameArray.Length == 2
+                    ? dnsNameArray
+                        .Last()
+                        .Replace("-", "_")
+                    : null;
+
+                // sometimes cross talk is desired between namespaces, append those not matching workload
+                if (!string.IsNullOrWhiteSpace(serviceNs))
```
garyburgmann, line 455: right now we can implicitly infer this when `!endpoint.IsInWorkloadNamespace`, as the namespace is [conditionally applied](https://github.com/Azure/Bridge-To-Kubernetes/blob/51ea890699b14a3213045b7e72625aea9db170b2/src/library/Connect/WorkloadInformationProvider.cs#L466)

src/library/Connect/LocalEnvironmentManager.cs
```diff
@@ -436,12 +436,26 @@ private void _ReportProgress(EventLevel eventLevel, string message, params objec
...
+                {
+                    serviceName = $"{serviceName}_{serviceNs}";
```
garyburgmann, line 457: I chose to do it this way as this is the way we think about the services with respect to k8s dns. So it made sense to simply append the ns here rather than mess with all the other logic setting env vars and append it elsewhere

src/library/Connect/LocalEnvironmentManager.cs
```diff
@@ -436,12 +436,26 @@ private void _ReportProgress(EventLevel eventLevel, string message, params objec
...
+                    .Split(".");
+
+                var serviceName = dnsNameArray
                     .First()
-                    .Replace("-", "_")
-                    .Replace(".", "_");
+                    .Replace("-", "_");
+
+                // when !endpoint.IsInWorkloadNamespace
+                var serviceNs = dnsNameArray.Length == 2
```
hsubramanianaks, line 448: @garyburgmann Thanks for the PR, can you add some description to the PR why you need this change and if possible link to a GH issue. Thanks.

garyburgmann:
@microsoft-github-policy-service agree company="Go1"

garyburgmann:
built out the CLI locally last week and worked as intended

image

src/library/Connect/LocalEnvironmentManager.cs
hsubramanianaks: @garyburgmann I agree is there any changes when dns name is  abc.companyname.com ? in that case you will need Replace(".", "_") ? Just thinking out aloud. Thank you for this PR.

hsubramanianaks (approved):
looks good to me @garyburgmann. We will test it and confirm.

src/library/Connect/LocalEnvironmentManager.cs
garyburgmann: @hsubramanianaks `.First()` has been around since before my changes. So in your example the string is simply `abc` that we are acting on.

cxznmhdcxz:
Passed on all 3 platforms.

PR #232, opened 2023-04-03
Enable auto review using chatgpt.

Tatsinnit:
This PR enables the auto review of the PR(s) the only caveat is that if description in the PR will be too long the token issues of chatgpt will kick in which could error out the workflow but beside that this workflow should suffice the first hand auto review.

fyi and thanks @hsubramanianaks  - we can chat real quick abut this and might be get a OpenAI token for this setup.  Thanks heaps, ❤️☕️🥷

### DIFF
--- a/.github/workflows/chatgpt-review.yml
+++ b/.github/workflows/chatgpt-review.yml
@@ -1,0 +1,18 @@
+name: ChatGPT Review
+
+on:
+  pull_request_target:
+    types: [opened, reopened, edited, ready_for_review]
+    branches:
+      - 'main'
+
+jobs:
+  chatgpt-review:
+    name: ChatGPT Review
+    runs-on: ubuntu-latest
+    steps:
+    - uses: feiskyer/ChatGPT-Reviewer@v0.3
+      name: ChatGPT Review
+      env:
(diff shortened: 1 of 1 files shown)


Eneuman:
This is awesome :) I'll use Copilot X to create the PR text, and you use Chat GPT to review it. What a technology age we are living it :)

.github/workflows/chatgpt-review.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  chatgpt-review:
+    name: ChatGPT Review
+    runs-on: ubuntu-latest
+    steps:
+    - uses: feiskyer/ChatGPT-Reviewer@v0.3
+      name: ChatGPT Review
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
hsubramanianaks, line 17: @Tatsinnit Thanks for this awesome PR. I like the idea like we discussed the other day. Only catch is we need to add these tokens to the repository for this GH action to work.

hsubramanianaks (approved):
LGTM

PR #242, opened 2023-04-18
Live test for PR validations

hsubramanianaks:
This PR introduces ```live test``` for Pull Request validations for Bridge repo. This PR runs specifically for ubuntu and checks if the change in the PR looks good. This is just basic live testing, will be added more on top of this.


.github/workflows/live-test.yml
```diff
@@ -0,0 +1,53 @@
+name: Live tests for B2K
+on: [push]
```
sabbour, line 2: On all push or on pull requests?

hsubramanianaks: @sabbour Just changed them :), it was from my fork which got missed. thank you.

hsubramanianaks: @sabbour on all pull requests. :)

.github/workflows/live-test.yml
```diff
@@ -0,0 +1,57 @@
...
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+jobs:
+  liveTest:
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
```
OliverMKing, line 11: Add macos or windows?

.github/workflows/live-test.yml
```diff
@@ -0,0 +1,57 @@
...
+      uses: medyagh/setup-minikube@master
+      with:
+       driver: docker
+    - name: Try the cluster !
+      run: kubectl get pods -A
+    - name: create todo-app namespace
+      run: kubectl create namespace todo-app
+    - name: Deploy todo-app
+      run:
+        kubectl apply -f https://raw.githubusercontent.com/Azure/Bridge-To-Kubernetes/main/samples/todo-app/deployment.yaml -n todo-app
```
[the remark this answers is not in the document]
hsubramanianaks: Thanks @OliverMKing this make sense. I missed this totally. :)

.github/workflows/live-test.yml
```diff
@@ -0,0 +1,57 @@
...
+      if: ${{ matrix.os == 'macos-latest' }}
+      uses: azure/setup-kubectl@v3
+      with:
+       version: latest
+    - name: Start minikube
+      uses: medyagh/setup-minikube@master
+      with:
+       driver: docker
+    - name: Try the cluster !
+      run: kubectl get pods -A
```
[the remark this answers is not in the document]
hsubramanianaks: @OliverMKing yes gives some time for cluster boot up, I think it is simple command that can be present. :) Please let me know.

.github/workflows/live-test.yml
```diff
@@ -0,0 +1,57 @@
...
+    name: live test for b2k
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup kubectl for macos-latest
+      if: ${{ matrix.os == 'macos-latest' }}
+      uses: azure/setup-kubectl@v3
+      with:
+       version: latest
+    - name: Start minikube
+      uses: medyagh/setup-minikube@master
```
OliverMKing, line 22: we should pin this to a sha instead of master. Master is a big security vulnerability.

.github/workflows/live-test.yml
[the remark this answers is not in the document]
OliverMKing: How does this give the cluster time to boot up? Doesn't that happen within the `medyagh/setup-minikube` action?

.github/workflows/live-test.yml
OliverMKing: Makes sense to me. That stuff is for sure a pain to deal with.

.github/workflows/live-test.yml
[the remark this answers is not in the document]
hsubramanianaks: @OliverMKing I mean this step make sures cluster is setup correctly. Just to display that minikube/k8s pods are running , doesn't take much time though :).  yes action ```medyagh/setup-minikube``` takes care of setting up minikube cluster. 

image

Tatsinnit (approved):
🎉☕️👍 nice one, thank you for the ping, the only thing notice is that add the sha pinning is required, here is a good link you can see for why and things https://github.com/marketplace/actions/ensure-sha-pinned-actions - I am not saying add this action into current workflow but the sentiment attached for the reasons. Rest looks super ☕️🙏

hsubramanianaks:
> 🎉☕️👍 nice one, thank you for the ping, the only thing notice is that add the sha pinning is required, here is a good link you can see for why and things https://github.com/marketplace/actions/ensure-sha-pinned-actions - I am not saying add this action into current workflow but the sentiment attached for the reasons. Rest looks super ☕️🙏

I agree we can add this action to whole repo so if anyone pushes a workflow file with uses: and without @sha will fail. Good one, thanks for the share. @Tatsinnit

.github/workflows/live-test.yml
hsubramanianaks: done updated to sha. good suggestion, thank you.

scripts/live_test.sh
```diff
@@ -0,0 +1,158 @@
...
+    start_minikube_tunnel
+    #set bridge as dev environment to run live test because some of image tags might noe be available in production
+    export BRIDGE_ENVIRONMENT='dev'
+    start_b2k
+
+    validate_b2k_is_running
+
+    stop_b2k
+
+    #ensure_b2k_is_disconnected
```
sabbour, line 145: Why is this one commented?

scripts/verify.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+# Verify if B2K is disconnected successfully
```
sabbour, line 2: As far as I understand, the entire script will exit with any failure. Is there value in using something like https://github.com/torokmark/assert.sh to run the tests in a more modular way?

scripts/live_test.sh
hsubramanianaks: @sabbour I tried adding them as single script but didn't work out so commented them. I will keep them for reference and clean up in next PR.

scripts/verify.sh
hsubramanianaks: @sabbour thanks for the suggestion. I will check and get back to you, I believe assert.sh are similar to exit codes.

scripts/live_test.sh
hsubramanianaks: By the way I took care of this test as another step in Github action (verify.sh as separate script)
image

hsubramanianaks: This will be cleaned up in upcoming PR. Thanks.

OliverMKing (approved):

PR #251, opened 2023-04-25
Ignore warning CS8002 when building in debug mode

### LINKED ISSUE #250: Warning CS8002 when building in debug mode
When building the projects in debug mode, you get alot of warnings that some reference assemblies do not have a strong name.

image

Theese warnings are realy not a problems since signing will be done when building a release, but they do make it alot harder to spot other warnings that should be fixed.

I suggest turning of the CS8002 rule for debug builds.

Eneuman:
This PR resolved 26 warnings by ignores the rule CS8002 when building in debug mode.

The rule CS8002 ("Assembly does not have a strong name") occures because signing is enabled on the projects.
Signing will be done in release build so I think this warning can be safely ignored in debug builds.

Fixes #250 

<strike> ```NoWarns for rules 1701;1702 are beeing added because they are default ignored in certain projects and when overriding the default behavior they get automaticly added.``` <strike>


src/EndpointManagerLauncher/endpointmanagerlauncher.csproj
```diff
@@ -19,6 +19,14 @@
     <ApplicationIcon>..\resources\icons\BridgeToKubernetes_VS_NoHalo.ico</ApplicationIcon>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>1701;1702;8002</NoWarn>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
```
hsubramanianaks, line 26: @Eneuman I found many links that is asking for supressing CS1701, CS1702 warnings from dotnet compilers. But I couldn't find the exact issue or link for it. Can you paste the link as comment here so that in future someone will know it. Thanks.

src/common/common.csproj
```diff
@@ -32,6 +32,12 @@
   <PropertyGroup>
     <DefineConstants>$(DefineConstants);$(TelemetryType)</DefineConstants>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net6.0|AnyCPU'">
+    <NoWarn>8002</NoWarn>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Release|net6.0|AnyCPU'">
```
hsubramanianaks, line 38: @Eneuman do we need this empty No warn with no codes in it ? May be this was missed.. :)

src/devhostagent.restorationjob/devhostAgent.restorationjob.csproj
```diff
@@ -16,6 +16,14 @@
     <HighEntropyVA>True</HighEntropyVA>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>8002</NoWarn>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
```
hsubramanianaks, line 23: @Eneuman same here empty No warn.

src/devhostagent/devhostAgent.csproj
```diff
@@ -17,6 +17,14 @@
     <HighEntropyVA>True</HighEntropyVA>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>8002</NoWarn>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
```
hsubramanianaks, line 24: same here @Eneuman

src/dsc/dsc.csproj
```diff
@@ -75,6 +75,14 @@
     <EndpointManagerSourcePath>$(EndpointManagerPublishDirectory)\$(EndpointManagerFileName)</EndpointManagerSourcePath>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>8002</NoWarn>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
+    <NoWarn></NoWarn>
```
hsubramanianaks, line 83: same here @Eneuman. Sorry about the multiple comments.

src/endpointmanager/endpointmanager.csproj
```diff
@@ -25,6 +25,14 @@
     <ApplicationIcon>..\resources\icons\BridgeToKubernetes_VS_NoHalo.ico</ApplicationIcon>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>8002</NoWarn>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
```
hsubramanianaks, line 32: same here @Eneuman

hsubramanianaks (changes requested):
looks good except the empty ```NO warn without codes``` and left a minor comment.

src/EndpointManagerLauncher/endpointmanagerlauncher.csproj
Eneuman: @hsubramanianaks I assumed that this was the default behavior because of how VS behaves.
If you remove the property groups so that the .csproj file does not have a <NoWarn> tag, VS GUI will look like this:

image

To me this indicates that for that specific project type, it will ignore CS1701, CS1702 by default, but maby thats a false asumption?

src/common/common.csproj
hsubramanianaks: That make sense :).

src/EndpointManagerLauncher/endpointmanagerlauncher.csproj
hsubramanianaks: @Eneuman Nope I agree VS GUI does that. But my point was is there GH issue or link why we are ignoring these warning for codes 1701,1702. if there is one please add them as comment in this file ?

Eneuman: I coudn't find any issue or link describing this VS behaviour so I reverted this changed :)

Eneuman:
I also removed the empty <NoWarns>.

hsubramanianaks (approved):
looks good.

PR #253, opened 2023-04-25
Fixes warnings for nullable reference types being used

### LINKED ISSUE #252: Invalid use of nullable reference type
Nullable reference types are being used in project that does not have this feature enabled.
This causes warnings on build.

image

Eneuman:
This PR fixes three occurences where nullable reference types where being used but the project had not enabled this feature.

Fixes #252


hsubramanianaks (approved):
@Eneuman Thanks looks good to me.

PR #261, opened 2023-04-27
Fixed exception rethrow

### LINKED ISSUE #260: Invalid stacktrace logged if a exception occurs when getting K8SConfiguration
If a exception occurs when reading K8SConfiguration, the stacktrace will not be written to logfile.

Eneuman:
This PR fixes a problem with a exception that was not rethrown correctly.
Now the stacktrace will be correct when it gets logged to file.

Fixed #260


hsubramanianaks (approved):
looks good to me. Thanks @Eneuman

PR #262, opened 2023-04-27
Fixed obsolete warnings and removed invalid assert

Eneuman:
This PR fixes 2 obsolete warnings and removes a invalid assert (elapsed can never be null).

### DIFF
--- a/src/library/Extensions/V1ContainerExtensions.cs
+++ b/src/library/Extensions/V1ContainerExtensions.cs
@@ -13,7 +13,8 @@ namespace k8s.Models
         {
             return StringComparer.OrdinalIgnoreCase.Equals(container.Name, "devspaces-proxy")
                     || StringComparer.OrdinalIgnoreCase.Equals(container.Name, "istio-proxy")
-                    || StringComparer.OrdinalIgnoreCase.Equals(container.Name, "daprd");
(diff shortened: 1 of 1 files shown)


src/dsc/Remoting/Startup.cs
```diff
@@ -35,7 +35,7 @@ public void ConfigureServices(IServiceCollection services)
         }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
-        public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory, IApplicationLifetime lifetime)
+        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
```
hsubramanianaks, line 38: @Eneuman IWebHostEnvironment env will still be unused do we want to keep it here or remove it if it not used ?

hsubramanianaks (approved):
looks good.

src/dsc/Remoting/Startup.cs
Eneuman: @hsubramanianaks Lets remove it since it unsused

cxznmhdcxz:
Test passed

hsubramanianaks (dismissed):

hsubramanianaks (approved):
looks good and tests are all passed.

PR #278, opened 2023-06-05
named ports fix for headless services

hsubramanianaks:
This PR is to support env variables for named ports for headless services. Named port were not set correctly due to previous PR which caused the named port not printing in right format , now it prints ex: HOST_NAME_SERVICE_NAME_SERVICE_PORT_NAMED_PORT_NAME
This will be helpful for users who rely on env variables especially for services using grpc.


src/library/Connect/WorkloadInformationProvider.cs
```diff
@@ -509,7 +509,7 @@ private int GetServicePortValue(V1ServicePort servicePort, V1Container container
                             DnsName = dns,
                             Ports = subset.Ports?
                                 .Where(port => this._IsSupportedProtocol(port.Protocol, endpoint.Metadata.Name) && !(portToIgnoreForHeadlessServiceEndpoints.GetValueOrDefault(endpoint.Metadata.Name)?.Contains(port.Port) ?? false))
-                                .Select(p => new PortPair(remotePort: p.Port, p.Name))
+                                .Select(p => new PortPair(remotePort: p.Port,name : p.Name))
```
hsubramanianaks, line 512: this was causing named port as empty or null which didn't set the env variable [here](https://github.com/Azure/Bridge-To-Kubernetes/blob/main/src/library/Connect/LocalEnvironmentManager.cs#LL501C21-L501C21)

cxznmhdcxz:
Passed on all 3 platforms.

peterbom (commented):
Looks good. Maybe in the future we could build some static `Create` methods for `PortPair` to ensure they're initialized in explicit valid ways, but for now I think it's good to just get the fix in.

peterbom (approved):

Tatsinnit (approved):
🎊🎊

PR #283, opened 2023-06-13
run certain workflows only if there are code changes in that PR/main 

hsubramanianaks:
This PR is to update certain workflow files to trigger only if there are source code or script file changes. PR's with read me changes or not source code changes will not trigger these workflows.


hsubramanianaks (commented):
leaving a comment to run required status checks

hsubramanianaks:
GitHub currently doesn't support easily skipping required status checks in the right way and we had to write custom code to make this efficient, seems like an overkill. So, closing this until GitHub supports this.

PR #263, opened 2023-04-27
WIP: .Net 7

Eneuman:
This PR updated all projects to .Net 7

Closes #228

PR #284, opened 2023-06-14
Dotnet 7 upgrade

hsubramanianaks:
This PR is to update CI/CD pipelines and GitHub actions to use dotnet 7. Thanks to @Eneuman for creating the initial PR# https://github.com/Azure/Bridge-To-Kubernetes/pull/263


Eneuman:
@hsubramanianaks I have noticed one issue after upgrading my local build bits to .Net 7, it might be just my computer but it's worth looking into.

In VS, if I start a debug session, stop it and then restarts it, and do this for a couple of times, I suddenly starts to receiving HTTP 401 from the SignalR communication.
The only way to get it working again is to kill the endpoint manage process.

The dev agent works just fine, the problem seems to be signalR in endpoint manager.

I can post some logs if you cannot reproduce it. Other than that it seems to be working fine.

hsubramanianaks:
> The dev agent works just fine, the problem seems to be signalR in endpoint manager.

Thanks for letting me know, I just started testing it. If I can't reproduce I will reach out for logs.

build-pipelines-PR.yml
```diff
@@ -194,6 +195,7 @@ jobs:
       restoreArguments: -r win-x64
       selectOrConfig: config
       nugetConfigPath: src/nuget.config
+      externalEndpoints: dba0ba50-2d4e-4f12-9f5a-42a638da803b
```
peterbom, line 198: Could you tell me what this guid refers to? What's an `externalEndpoint` here?

peterbom: From a quick read of the [docs](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/dotnet-core-cli-v2?view=azure-pipelines) it looks as if this is an alias for a [service connection](https://learn.microsoft.com/en-us/azure/devops/pipelines/library/service-endpoints?view=azure-devops&tabs=yaml) for a NuGet feed... Has this been working until now?

peterbom (approved):
Looks good to me. I just had one question about the `externalEndpoints` used in the pipeline `restore` commands. Have you been able to do test runs of the pipelines to confirm they run successfully?

hsubramanianaks:
> Looks good to me. I just had one question about the `externalEndpoints` used in the pipeline `restore` commands. Have you been able to do test runs of the pipelines to confirm they run successfully?

That is something I am trying to see if that fixes the issue of the build failing intermittently. Since Nuget feeds are external. I will read more about it and confirm here .

build-pipelines-PR.yml
hsubramanianaks: I haven't tested it, will confirm if it works. Thanks for the comment.

hsubramanianaks:
> @hsubramanianaks I have noticed one issue after upgrading my local build bits to .Net 7, it might be just my computer but it's worth looking into.
> 
> In VS, if I start a debug session, stop it and then restarts it, and do this for a couple of times, I suddenly starts to receiving HTTP 404 from the SignalR communication. The only way to get it working again is to kill the endpoint manage process.
> 
> The dev agent works just fine, the problem seems to be signalR in endpoint manager.
> 
> I can post some logs if you cannot reproduce it. Other than that it seems to be working fine.

@Eneuman I couldn't reproduce this, It works fine for me. Can you please share the logs ?

build-pipelines-PR.yml
hsubramanianaks: @peterbom Previous failing build - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7918956&view=logs&s=96ac2280-8cb4-5df5-99de-dd2da759617d

After adding external endpoint , passing build - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7919390&view=results

It is weird though why it doesn't fail if we provide external endpoint.

Eneuman:
@hsubramanianaks I was able to reporoduce this today. It happend after a few starts and stops.
My build is from latest in .Net 7 branch.

I have also tested a local build from latest master branch and that works as expected.
I think next step would be to enbale verbose logging for SignalR on both client and server side to see what happens in the handshake.

[bridge-mindarocli-2023-06-17-16-57-05-24180.txt](https://github.com/Azure/Bridge-To-Kubernetes/files/11779617/bridge-mindarocli-2023-06-17-16-57-05-24180.txt)

hsubramanianaks:
@Eneuman if this is the error 

2023-06-17T16:58:51.4818965Z | MindaroCli | TRACE | Port forward klinikportalservice-6576b656fb-h8fln 64835:50052 64890 : Write to remote failed with The server returned status code '404' when status code '101' was expected.

Are you sure devhost agent is running and not restoring prematurely ? This 404 happens when devhost agent is not running on port 50052 and when port forwarding to dev host agent fails. Restore happens if signalR connection from your local to devhost agent failed. I agree turning on verbose logs, May be add more logs to this file - 

https://github.com/Azure/Bridge-To-Kubernetes/blob/main/src/common/DevHostAgent/DevHostAgentExecutorClient.cs in method _connectAsync method where the connection happens. Thank you, please let me know how it goes.

hsubramanianaks:
> @Eneuman if this is the error
> 
> 2023-06-17T16:58:51.4818965Z | MindaroCli | TRACE | Port forward klinikportalservice-6576b656fb-h8fln 64835:50052 64890 : Write to remote failed with The server returned status code '404' when status code '101' was expected.
> 
> Are you sure devhost agent is running and not restoring prematurely ? This 404 happens when devhost agent is not running on port 50052 and when port forwarding to dev host agent fails. Restore happens if signalR connection from your local to devhost agent failed. I agree turning on verbose logs, May be add more logs to this file -
> 
> https://github.com/Azure/Bridge-To-Kubernetes/blob/main/src/common/DevHostAgent/DevHostAgentExecutorClient.cs in method _connectAsync method where the connection happens.
> 
> I can see from the logs _connectAsync signalR method failed with below error
> 
> **System.Net.Http.HttpRequestException: System.Net.Http.HttpRequestException: An error occurred while sending the request.\n ---> System.IO.IOException: The response ended prematurely.\n at System.Net.Http.HttpConnection.SendAsyncCore(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)\n --- End of inner exception stac**
> 
> Another option is to increase time to restore using env variables and see if it works. Ex: increase the time in restoration job here -
> 
> https://github.com/Azure/Bridge-To-Kubernetes/blob/main/src/devhostagent.restorationjob/RestorationJobEnvironmentVariables.cs _restoreTimeout variable is 1mi Ute now you can change it to 30 mins. Just a suggestion.
> 
> And build your own image and override it using env variables BRIDGE_RESTORATIONJOBIMAGENAME
> 
> Thank you, please let me know how it goes.

@Eneuman any update on the issue you are facing? my testing team didn't report any issues until now for VS. Please let us know. Thank you.

azure-pipelines.yml
```diff
@@ -202,12 +202,12 @@ jobs:
     displayName: Copy EndpointManager Launcher to dsc folder
     enabled: True
     inputs:
-      SourceFolder: src/EndpointManagerLauncher/bin/${{ parameters.BuildConfiguration }}/net6.0/win-x64/publish/
+      SourceFolder: src/EndpointManagerLauncher/bin/${{ parameters.BuildConfiguration }}/net7.0/win-x64/publish/
```
bfoley13, line 205: any reason we dont have `netx.x` as a param as well so we dont have to change it here every time?

bfoley13 (approved):
Mostly config changes and they look good to me! Would only say we might want to see if we can parameterize some of these values more to reduce the changes needed on the next upgrade!

azure-pipelines.yml
hsubramanianaks: @bfoley13 thanks for this good suggestion, I will check it out.

bfoley13 (approved):

Eneuman:
@hsubramanianaks I have an idea that it might be SignalR negotiation that creates this exception, so I'm going see if it gets resolved by setting transport to "websockets" and skipNegotation to true.

Today the code allows negotiation, is this realy nessasary? I cannot se "long polling" as an options in this senario so maby negotiation should be disabled anyway.

hsubramanianaks:
> .SendAsyncCore(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)\n --- End of inner exception stac

@Eneuman so basically you are looking at option of setting transport as websockets and skipNegotiation as true ? 
image

I would say you can try and see if it works. This has been not reproducible at our end, testing team tried multiple times to recreate it but they couldn't. how are you testing these changes ? can you give more specifics about it please ?

Eneuman:
@hsubramanianaks Today I was able to reproduce my error with the current released .Net 6 bits so I think the problem has nothing to do with .Net 7. I will continue to try and track down the issue but I don't think this is something that should stop this PR from progressing :)

hsubramanianaks:
> @hsubramanianaks Today I was able to reproduce my error with the current released .Net 6 bits so I think the problem has nothing to do with .Net 7. I will continue to try and track down the issue but I don't think this is something that should stop this PR from moving forward :)

@eneuman Yes I have seen this , happened rarely to me. Please create an issue with all details, we can track it down and see how to fix this problem. we will move on with .NET 7 upgrade. Thanks for the update.

Issue #223, opened 2023-03-29
Unable to connect because of xtables lock


SteveCurran:
**Describe Bug**
Unable to connect to pod because another process has an xtables lock

"Bridge to Kubernetes initialization (VS Code v1.76.2 - Extension v2.0.120230310)
Logs: /home/steve/.config/Code/logs/20230329T125341/window1/exthost/mindaro.mindaro/mindaro-vscode-2023-03-29T17-53-45.566Z.txt
Waiting for 'cspbatch-746df866f9-rpt74' in namespace 'k8sdev' to reach running state...
Deployment 'k8sdev/cspbatch' patched to run agent.
Remote agent deployed in container 'cspbatch' in pod 'cspbatch-746df866f9-rpt74'.
Preparing to run Bridge To Kubernetes configured as pod k8sdev/cspbatch-746df866f9-rpt74 ...
Connection established.
Running 'iptables' failed with exit code '4': 'Waiting for process 942159
Another app is currently holding the xtables lock. Perhaps you want to use the -w option?
Process has exited with exit code 4
'
Stopping workload and cleaning up...
Restore: Deployment 'cspbatch' restored to original state with image 'localhost:5000/batch:latest'.
Failed to establish a connection. Error: Running 'iptables' failed with exit code '4': 'Waiting for process 942159
Another app is currently holding the xtables lock. Perhaps you want to use the -w option?
Process has exited with exit code 4"

I am able to successfully connect after restarting the ubuntu vm.

**Mention the platform you are using**
I am running b2k on ubuntu 20.04, vscode, and local k3s

[mindaro-vscode-2023-03-29T17-53-45.566Z.txt](https://github.com/Azure/Bridge-To-Kubernetes/files/11104177/mindaro-vscode-2023-03-29T17-53-45.566Z.txt)

hsubramanianaks:
@SteveCurran Thanks for raising this issue. are you facing this problem every time you use Bridge, or it is intermittent? Can you check and send the output screenshot here if you see any process running for PID ```942159``` in your machine ? As far as i know when bridge trys to run the iptables command, it fails due to another app ex: firewalld has the xtables locked.

SteveCurran:
[...]

PR #289, opened 2023-07-06
Iptables wait option

hsubramanianaks:
This PR is to fix https://github.com/Azure/Bridge-To-Kubernetes/issues/223.

- This code adds a wait time of 30 seconds for iptables operation which would wait for adding iptables rules for redirection when xtables are locked by another process. In most cases this will not happen. Thank you.

### DIFF
--- a/src/common/IP/IPManager.cs
+++ b/src/common/IP/IPManager.cs
@@ -205,8 +205,9 @@ namespace Microsoft.BridgeToKubernetes.Common.IP
                     {
                         foreach (var portPair in endpoint.Ports)
                         {
-                            rules.Add($"--table nat --append PREROUTING -p tcp --dst {endpoint.LocalIP} --dport {portPair.RemotePort} --jump DNAT --to-destination {endpoint.LocalIP}:{portPair.LocalPort}");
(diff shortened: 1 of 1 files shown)


peterbom (approved):
Out of interest, what is the behaviour of `iptables` if you don't specify the `-w` parameter? It fails immediately?

Vidya2606 (dismissed):

hsubramanianaks:
yes it would fail immediately default is 1 second.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

src/common/IP/IPManager.cs
```diff
@@ -297,7 +298,7 @@ private void RemoveRoutingRules(CancellationToken cancellationToken, IPAddress[]
                 command: args,
                 logCallback: (line) => _log.Info(line),
                 envVariables: null,
-                timeout: TimeSpan.FromSeconds(10),
+                timeout: TimeSpan.FromSeconds(60),
```
Tatsinnit, line 301: 💡minor suggestion- if we comment this line with same rationale as mention in description so anyone reading will have relevance latter.

Tatsinnit (approved):
👍💭

src/common/IP/IPManager.cs
hsubramanianaks: @Tatsinnit Thanks for the suggestion, done.

Tatsinnit (approved):

Issue #140, opened 2023-01-05
Bridge End to End Integration Tests


elenavillamil:
Today Bridge only has UTs. All integration test is done manually. With this feature we will add integration test coverage that is automated and runs with every PR. 

General idea of how this can be accomplished: Set up a pipeline that runs following tasks:
* Create test cluster
* Deploy todo app
* Connect bridge to stats-api, and verify connect happened successfully
* Add todo item using frontend url
* Go to stats-api link and validate correct entries returned.

hsubramanianaks:
I like this idea; it reduces the load on testing team. I was talking about this with Joyer last night. Suggested looking for tools like cypress, playwright to automate it or write a script to test them as CLI. there are several ways to accomplish this. Thank you for raising this enhancement request. @GatoonJin  FYI.

elenavillamil:
@hsubramanianaks assigned this to you since you started looking into it. Look forward to see what you come up with :)

hsubramanianaks:
we have live tests here for linux, similar needs to be done with windows and macos. One restriction is installing minikube or kind clusters in windows/macos runner are tough and causes issues. Only other way is to create an AKS cluster under testing subscription and use that cluster for live automated tests. @cxznmhdcxz FYI.

hsubramanianaks:
@cxznmhdcxz As discussed can you add more complex tests to these live tests and make them parallel? Today it checks out the projects, setups cluster and runs b2k connect (basic testing). But do we want to enhance this live testing experience to more use cases like test routing, ingress, managed identity, dapr etc.. Thank you.

hsubramanianaks:
Closing this as this was completed by Ruijun.

PR #302, opened 2023-08-07
add live-test dispatch workflow for windows and mac

cxznmhdcxz:
See https://github.com/Azure/Bridge-To-Kubernetes/issues/140
This PR: 
Adds live test for windows and mac, workflow will create an AKS cluster to run the test.
No changes to test for linux
Now the tests for all 3 platforms will be triggered by slash command and `sha` input like `/ok-to-test sha=1234567` in PR comment

A latest run in my fork: https://github.com/cxznmhdcxz/Bridge-To-Kubernetes/actions/runs/5798264426

### DIFF
--- a/.github/workflows/live-test.yml
+++ b/.github/workflows/live-test.yml
@@ -8,8 +8,7 @@ jobs:
       PR_NUMBER: ${{ github.event.client_payload.slash_command.args.named.PR_NUMBER }}
     strategy:
       matrix:
-        # os: [ubuntu-latest, windows-latest, macos-latest]
-         os: [macos-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
     runs-on: ${{ matrix.os }}
     name: live test for b2k
     steps:
@@ -24,34 +23,28 @@ jobs:
       run: |
         echo "mac_cluster=b2k-livetest-mac-pr-$PR_NUMBER" >> $GITHUB_ENV
         echo "windows_cluster=b2k-livetest-windows-pr-$PR_NUMBER" >> $GITHUB_ENV
-    # todo: actually create an AKS
+    # create cluster for Windows
     - name: Create AKS for Windows
       if: ${{ matrix.os == 'windows-latest' }}
-      uses: Azure/powershell@v1
-      with:
-        inlineScript: |
-          az resource list -g ${{ secrets.AZURE_RESOURCE_GROUP }}
-        azPSVersion: '3.1.0'
+      run: |
+        az aks create -g ${{ secrets.AZURE_RESOURCE_GROUP }} -n ${{ env.windows_cluster }} --no-ssh-key
     - name: Set AKS Context for Windows
       if: ${{ matrix.os == 'windows-latest' }}
       uses: azure/aks-set-context@v3
       with:
-        # todo: set to the actually created AKS
         resource-group: ${{ secrets.AZURE_RESOURCE_GROUP }}
-        cluster-name: testworkflow-aks
-    - name: Create AKS Mac
+        cluster-name: ${{ env.windows_cluster }}
+    # create cluster for Mac
+    - name: Create AKS mac
       if: ${{ matrix.os == 'macos-latest' }}
-      uses: Azure/powershell@v1
-      with:
-        inlineScript: |
-          az aks create -g ${{ secrets.AZURE_RESOURCE_GROUP }} -n $mac_cluster
-        azPSVersion: '3.1.0'
+      run: |
+        az aks create -g ${{ secrets.AZURE_RESOURCE_GROUP }} -n ${{ env.mac_cluster }} --no-ssh-key
     - name: Set AKS Context for MacOS
       if: ${{ matrix.os == 'macos-latest' }}
       uses: azure/aks-set-context@v3
       with:
         resource-group: ${{ secrets.AZURE_RESOURCE_GROUP }}
-        cluster-name: $mac_cluster
+        cluster-name: ${{ env.mac_cluster }}
     - name: Setup kubectl for macos-latest
       if: ${{ matrix.os == 'macos-latest' }}
       uses: azure/setup-kubectl@v3
--- a/.github/workflows/live-test.yml
+++ b/.github/workflows/live-test.yml
@@ -2,6 +2,8 @@ name: Live tests for B2K
 on:
   repository_dispatch:
     types: [ok-to-test-command]
+permissions:
+  id-token: write
 jobs:
   liveTest:
     env:
@@ -10,7 +12,7 @@ jobs:
       HEAD_BRANCH: ${{ github.event.client_payload.slash_command.args.named.HEAD_BRANCH }}
     strategy:
       matrix:
-        os: [macos-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     name: live test for b2k
     steps:
@@ -23,7 +25,9 @@ jobs:
       if: ${{ matrix.os != 'ubuntu-latest' }}
       uses: Azure/login@v1
       with:
-        creds: '{"clientId":"${{ secrets.AZURE_CLIENT_ID }}","clientSecret":"${{ secrets.AZURE_CLIENT_SECRET }}","subscriptionId":"${{ secrets.AZURE_SUBSCRIPTION_ID }}","tenantId":"${{ secrets.AZURE_TENANT_ID }}"}'
+        tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+        subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+        client-id: ${{ secrets.AZURE_CLIENT_ID }}
     - name: Prepare Cluster name Windows
       if: ${{ matrix.os == 'windows-latest' }}
       run: |
--- a/.github/workflows/live-test.yml
+++ b/.github/workflows/live-test.yml
@@ -6,13 +6,19 @@ jobs:
   liveTest:
     env:
       PR_NUMBER: ${{ github.event.client_payload.slash_command.args.named.PR_NUMBER }}
+      HEAD_REPO: ${{ github.event.client_payload.slash_command.args.named.HEAD_REPO }}
+      HEAD_BRANCH: ${{ github.event.client_payload.slash_command.args.named.HEAD_BRANCH }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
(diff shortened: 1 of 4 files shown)


cxznmhdcxz:
@hsubramanianaks  please review thanks

.github/workflows/live-test.yml
```diff
@@ -1,24 +1,70 @@
...
-  pull_request:
-    branches: [ "main" ]
+  repository_dispatch:
+    types: [ok-to-test-command]
+permissions:
+  id-token: write
 jobs:
   liveTest:
+    env:
+      PR_NUMBER: ${{ github.event.client_payload.slash_command.args.named.PR_NUMBER }}
```
hsubramanianaks, line 10: @cxznmhdcxz This env is from /ok-to-test command so that user doesn't have to specify the sha argument and it always get's the latest sha from the branch and runs it correct ? Just adding here as FYI for others to know.

.github/workflows/live-test.yml
```diff
@@ -1,24 +1,70 @@
...
     - uses: actions/checkout@v2
+      # checkout PR head repo/branch
+      with:
+         repository: ${{ env.HEAD_REPO }}
+         ref: ${{ env.HEAD_BRANCH }}
+    - name: Login Azure Context
+      if: ${{ matrix.os != 'ubuntu-latest' }}
+      uses: Azure/login@v1
+      with:
+        tenant-id: ${{ secrets.AZURE_TENANT_ID }}
```
hsubramanianaks, line 28: @cxznmhdcxz FYI - these secrets tenantID, SubscriptionID and CLIENT ID has to be added as GH secrets for this workflow to run smoothly.

.github/workflows/live-test.yml
```diff
@@ -1,24 +1,70 @@
...
+        subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+        client-id: ${{ secrets.AZURE_CLIENT_ID }}
+    - name: Prepare Cluster name Windows
```
hsubramanianaks, line 31: @cxznmhdcxz Can we combine the prepare and create AKS cluster steps for mac and windows since they are common using variable ```$RUNNER_OS```? for example 

```
- name: Prepare Cluster name Mac
      if: ${{ matrix.os == 'macos-latest' ||  matrix.os == 'windows-latest' }}
      shell: bash
      run: |
        echo "cluster_name=b2k-livetest-$RUNNER_OS-pr-$PR_NUMBER" >> $GITHUB_ENV
- name: Create AKS mac
      if: ${{ matrix.os == 'macos-latest' ||  matrix.os == 'windows-latest' }}
      run: |
        az aks create -g ${{ secrets.AZURE_RESOURCE_GROUP }} -n ${{ env.cluster_name }} --no-ssh-key

```
so that it will be 2 steps instead of 4 separate steps.

scripts/verify.sh
```diff
@@ -30,7 +30,7 @@ verify_if_multiple_pods_exists() {
     echo "Verifying if multiple pods exists"
     count=0
     MULTIPLE_POD_EXISTS=false
-    while [ "$count" -le 3 ]; do
+    while [ "$count" -le 10 ]; do
```
hsubramanianaks, line 33: @cxznmhdcxz Please add the reason why we are increasing the count from 3 to 10 here. Thanks.

hsubramanianaks (changes requested):
@cxznmhdcxz Looks great, good addition to existing live tests. Really appreciate for making these changes. I would recommend looking into my comments (minor one's) and resolve them once the changes are done. Also, to the PR description, please add screen shots or links to GH actions runs from your fork that it works without any issues or errors would be helpful for others to review this PR. Thank you :)

.github/workflows/live-test.yml
```diff
@@ -1,24 +1,70 @@
...
+    - name: Create AKS mac
+      if: ${{ matrix.os == 'macos-latest' }}
+      run: |
+        az aks create -g ${{ secrets.AZURE_RESOURCE_GROUP }} -n ${{ env.mac_cluster }} --no-ssh-key
+    - name: Set AKS Context for MacOS
+      if: ${{ matrix.os == 'macos-latest' }}
+      uses: azure/aks-set-context@v3
+      with:
+        resource-group: ${{ secrets.AZURE_RESOURCE_GROUP }}
+        cluster-name: ${{ env.mac_cluster }}
```
hsubramanianaks, line 60: @cxznmhdcxz Can we clean up these AKS clusters after B2K is disconnected or as last step so that we don't use up more resources ? example : ```az aks delete -n cluster-name -g resource-group-name -y --no-wait ``` ?

.github/workflows/ok-to-test.yml
```diff
@@ -0,0 +1,55 @@
...
+        uses: peter-evans/slash-command-dispatch@a28ee6cd74d5200f99e247ebc7b365c03ae0ef3c # v3.0.1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          reaction-token: ${{ secrets.GITHUB_TOKEN }}
+          issue-type: pull-request
+          commands: ok-to-test
+          static-args: |
+            PR_NUMBER=${{ github.event.issue.number }}
+            HEAD_REPO=${{ fromJSON(steps.resolve-pr.outputs.result).head.repo.full_name }}
+            HEAD_BRANCH=${{ fromJSON(steps.resolve-pr.outputs.result).head.ref }}
```
OliverMKing, line 52: setting the ref that is checked out this way opens us up to security vulnerabilities. We can use the event payload to avoid that `github.event.client_payload.slash_command.args.named.sha`.

Without this an approver can mark a pr as `/ok-to-test` then an attacker can push a malicious commit to the PR that will be run. The purpose of the `/ok-to-test` is that an approver can approve an exact set of changes to ensure there's nothing malicious run with our secrets. Using the sha method makes that attack impossible. https://github.com/Azure/aks-app-routing-operator/blob/1028b62d138483de6dfca64c54ed39c11c54c50f/.github/workflows/pr-validate-fork.yaml#L42 Here's an example of how we do this in another repo. We have quite a few checks against the sha.

.github/workflows/live-test.yml
OliverMKing: I left a comment on this as well. Without the sha this opens us up to a security vulnerability. Good callout!

.github/workflows/live-test.yml
```diff
@@ -32,7 +78,7 @@ jobs:
         kubectl wait --for=condition=ready pod --all -n todo-app --timeout=300s
         kubectl get pods -n todo-app
     - name: Publish and test
-      timeout-minutes: 10
+      timeout-minutes: 30
```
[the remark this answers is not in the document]
cxznmhdcxz: The run on Windows might time out if it's set to 10 minutes.
Like this run:
https://github.com/cxznmhdcxz/Bridge-To-Kubernetes/actions/runs/5726686544/job/15518332915?pr=4

.github/workflows/ok-to-test.yml
hsubramanianaks: @OliverMKing so you are suggesting to add the following if conditions in the workflow and so approver has to specify which sha to use for the workflows correct ? ex: ```/ok-to-test sha=123456```. But why do you say without specifying sha it would cause security vulnerabilities ?

```
if:
      github.event_name == 'repository_dispatch' &&
      github.event.client_payload.slash_command.args.named.sha != '' &&
      contains(github.event.client_payload.pull_request.head.sha, github.event.client_payload.slash_command.args.named.sha)
```

 @cxznmhdcxz made changes to get the latest sha for that run and comment /ok-to-test when approver has "write" access only for that one time run. Approver has to make sure the latest sha commit doesn't have malicious code. Please correct If my understanding is wrong.

hsubramanianaks: @OliverMKing Thanks for explaining, it would be a very short span of time though we can't compromise security for it. @cxznmhdcxz Ruijun this needs to be changed back accept ```/ok-to-test sha=123456``` with sha arguments. Please let us know if you have any questions or concerns.

.github/workflows/live-test.yml
```diff
@@ -1,24 +1,61 @@
...
+        client-id: ${{ secrets.AZURE_CLIENT_ID }}
+    - name: Prepare Cluster name Mac and Windows
+      if: ${{ matrix.os == 'macos-latest' ||  matrix.os == 'windows-latest' }}
+      shell: bash
+      run: |
+        if [ "$RUNNER_OS" == "Windows" ]; then
+          echo "cluster_name=b2k-livetest-win-pr-$PR_NUMBER" >> $GITHUB_ENV
+        else
+          echo "cluster_name=b2k-livetest-mac-pr-$PR_NUMBER" >> $GITHUB_ENV
+        fi
```
cxznmhdcxz, line 39: @hsubramanianaks thanks for suggestion. Combined AKS steps but still need to do this here because "windows" is not allowed as either name or substring FYI.

hsubramanianaks: Yep agree, cluster name is required in the environment variables since it is used in subsequent steps.

.github/workflows/live-test.yml
hsubramanianaks: this is updated to input the commit sha.

.github/workflows/live-test.yml
hsubramanianaks: These are added to main repo secrets, thanks for the help @cxznmhdcxz .

hsubramanianaks (approved):
Looks good, Thanks for doing this PR @cxznmhdcxz.

hsubramanianaks (changes requested):
Sorry found some last minute changes, requesting to change them. @cxznmhdcxz

Tatsinnit (commented):
❤️☕️💡 Thank you so much for this, the work looks super and awesome! 

I have a question:

What is the real scenario for this? I might not questioning this work just understanding the user case: `We will now trigger tests only using this tag?` instead of always dispatch run for `main` branch? and we are doing this because reproduces test consume is a lot heavy? (If so lets talk more around that as well please) 

Thanks heaps.

Tatsinnit:
> ❤️☕️💡 Thank you so much for this, the work looks super and awesome!
> 
> I have a question:
> 
> What is the real scenario for this? I might not questioning this work just understanding the user case: `We will now trigger tests only using this tag?` instead of always dispatch run for `main` branch? and we are doing this because reproduces test consume is a lot heavy? (If so lets talk more around that as well please)
> 
> Thanks heaps.

Ok, I know now, the background et. al. Thank you

Tatsinnit (approved):

PR #304, opened 2023-08-11
updating image tags

hsubramanianaks:
This PR is to update image tags to update CBL mariner to latest images which includes several fixes , refer [here](https://github.com/microsoft/CBL-Mariner/releases)

note: There are no tags or image name updates, it just pulling the latest CBL mariner images from MCR for sdk and asp.net images using docker build and push them to ACR. The tag updates for B2K images (devhost, restoration, routing manager) is required because users running with older images will not get this update unless there is tag change. Following this there will releases for vscode and vs extensions. Thank you.


bfoley13 (approved):

gambtho (approved):

Tatsinnit (approved):
☕️🤙

PR #310, opened 2023-09-05
add linkerd to supported sidecars

hsubramanianaks:
This PR is to add support for side cars like linkerd, nginx for bridge. IsKnownSidecarContainer method would filter out containers running with same name and will select the actual user's workload containers.

### DIFF
--- a/src/library/Extensions/V1ContainerExtensions.cs
+++ b/src/library/Extensions/V1ContainerExtensions.cs
@@ -15,10 +15,10 @@ namespace k8s.Models
             var excludedContainerNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase) {
                 "istio-proxy",
                 "linkerd-proxy",
-                "envoy",
-                "istio-init",
-                "linkerd-init",
-                "istio-egressgateway",
+                "devspaces-proxy",
+                "nginx-proxy",
+                "jaeger-agent",
+                "daprd"
             };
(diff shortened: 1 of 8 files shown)


OliverMKing (approved):

src/library/Extensions/V1ContainerExtensions.cs
```diff
@@ -14,7 +14,9 @@ public static bool IsKnownSidecarContainer(this V1Container container)
             return StringComparer.OrdinalIgnoreCase.Equals(container.Name, "devspaces-proxy")
                     || StringComparer.OrdinalIgnoreCase.Equals(container.Name, "istio-proxy")
                     || StringComparer.OrdinalIgnoreCase.Equals(container.Name, "daprd")
-                    || StringComparer.OrdinalIgnoreCase.Equals(container.Name, "jaeger-agent");
+                    || StringComparer.OrdinalIgnoreCase.Equals(container.Name, "jaeger-agent")
+                    || StringComparer.OrdinalIgnoreCase.Equals(container.Name, "linkerd-proxy")
+                    || StringComparer.OrdinalIgnoreCase.Equals(container.Name, "nginx-proxy");
```
OliverMKing, line 19: Maybe slap these in an array sometime so there's less duplicated code? You could iterate over the known sidecar array and perform the check on each.

OliverMKing (approved):

bfoley13 (approved):

PR #316, opened 2023-09-13
update managed identity clientid to latest

hsubramanianaks:
This PR is to update managed identity client id to right one since manual testing was failing and previous PR missed updating this.

### DIFF
--- a/samples/managed-identity/KubernetesLocalProcessConfig.yaml
+++ b/samples/managed-identity/KubernetesLocalProcessConfig.yaml
@@ -5,6 +5,6 @@ env:
   - name: STORAGE_CONTAINER_NAME
     value: "mitestsa-container"
   - name: MI_CLIENT_ID
-    value: "bb502719-354c-4ca3-8991-0bff00ffc7a5"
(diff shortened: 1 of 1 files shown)


Vidya2606 (approved):

samples/managed-identity/KubernetesLocalProcessConfig.yaml
```diff
@@ -5,6 +5,6 @@ env:
   - name: STORAGE_CONTAINER_NAME
     value: "mitestsa-container"
   - name: MI_CLIENT_ID
-    value: "bb502719-354c-4ca3-8991-0bff00ffc7a5"
+    value: "b24d4637-3453-4f6f-8187-228a5442a675"
```
Tatsinnit, line 8: 💡quick question - what for a this guid signify?

hsubramanianaks: It's managed identity client id for the sample app which is used for manual E2E testing. Recently I recreated the clusters and managed identity which are used for testing for S360 warnings so I had to change the client id. Hopefully this makes it clear.

PR #338, opened 2023-11-14
Fix for restoration job to find agent endpoints when there are multiple replicas/stateful sets.

### LINKED ISSUE #337: Restoration Job fails to get agent endpoint when there are multiple replicas for a service/pod
**Is your feature request related to a problem? Please describe.**
Restoration Job fails to get agent endpoint when there are multiple replica sets for a service/pod 

**Mention what platform you want to support the new feature**
all platforms

**Describe the solution you'd like**
Restoration job need to identify the pod and get the agent endpoint successfully. 
Pass the selected pod name while creating restoration job in a config map which would save a http call to k8s api to find out the pod name.

hsubramanianaks:
This PR handles the following.

- Fixes #337 
- remove the use of dynamic (causes performance bottleneck) and replace it with switch case. 
- restructure the restoration job class to handle multiple replicas/stateful sets. 
- More unit test cases.


deployment/settings/services/imagetag.setting
```diff
@@ -1,4 +1,4 @@
 MINDARO_DEVHOSTAGENT_TAG=1.3.4
-MINDARO_DEVHOSTAGENT_RESTORATIONJOB_TAG=1.3.4
+MINDARO_DEVHOSTAGENT_RESTORATIONJOB_TAG=issue337
```
hsubramanianaks, line 2: this is temporary changes for testing , I will revert them back to right versions.

Issue #234, opened 2023-04-03
Bridge is crashed if UDP svc at namespace. (Index was outside the bounds error)


Rverkh:
**Describe the bug**
Bridge doesn't work if we have UDP service (svc) at namespace where we are trying to replace application with Bridge.
An unexpected error occurred: 'Index was outside the bounds of the array.'

As far as I know kubectl doesn't support UDP port-forward yet. But may be it's possible to just ignore UDP svc and write some warnings at that case?

**Mention the platform you are using**
Reproduced with vscode extension and cli tool

**To Reproduce**
1. Deploy tcp based application
2. Deploy any udp based application at the same namespace
3. Make svc's for both
4. Launch b2k for tcp application and it will fault with "An unexpected error occurred: 'Index was outside the bounds of the array.'" error

Here is some simple example of pod and svc
```
apiVersion: v1
kind: Pod
metadata:
  name: sample-udp-listener
  labels:
    name: sample-udp-listener
  namespace: udp-example
spec:
  containers:
    - name: "sample-udp-listener"
      image: "alpine"
      command: ["/usr/bin/nc"]
      args: [ "-u", "-l","2115"]
      ports:
        - containerPort: 2115
          protocol: UDP
---
apiVersion: v1
kind: Pod
metadata:
  name: sample-tcp-listener
  labels:
    name: sample-tcp-listener
  namespace: udp-example
spec:
  containers:
    - name: "sample-tcp-listener"
      image: "alpine"
      command: ["/usr/bin/nc"]
      args: [ "-lk", "-e","2116"]
      ports:
        - containerPort: 2116
          protocol: TCP
---
apiVersion: v1
kind: Service
metadata:
  labels:
    name: sample-udp-svc
  name: sample-udp-svc
  namespace: udp-example
spec:
  ports:
  - port: 2115
    protocol: UDP
    targetPort: 2115
  selector:
    name: sample-udp-listener
  type: ClusterIP
---
apiVersion: v1
kind: Service
metadata:
  labels:
    name: sample-tcp-svc
  name: sample-tcp-svc
  namespace: udp-example
spec:
  ports:
  - port: 2116
    protocol: TCP
    targetPort: 2116
  selector:
    name: sample-tcp-listener
  type: ClusterIP
```

Then you can launch 

```
./dsc connect --service sample-tcp-svc --local-port 3116  --namespace udp-example --use-kubernetes-service-environment-variables


[...]

PR #339, opened 2023-11-26
Skip endpoint without ports in LocalEnvironmentManager

ababilone:
Fix https://github.com/Azure/Bridge-To-Kubernetes/issues/234

This PR fixes Index out of bound when user's namespace contains UDP services without ports.

### DIFF
--- a/src/library/Connect/LocalEnvironmentManager.cs
+++ b/src/library/Connect/LocalEnvironmentManager.cs
@@ -426,7 +426,7 @@ namespace Microsoft.BridgeToKubernetes.Library.Connect
 
             foreach (var endpoint in workloadInfo.ReachableEndpoints)
             {
-                if (endpoint.Ports.Length == 0)
+                if (endpoint.Ports == null || endpoint.Ports.Length == 0)
                 {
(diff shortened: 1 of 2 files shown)


ababilone:

src/library/Connect/LocalEnvironmentManager.cs
```diff
@@ -426,6 +426,12 @@ private void _ReportProgress(EventLevel eventLevel, string message, params objec
 
             foreach (var endpoint in workloadInfo.ReachableEndpoints)
             {
+                if (endpoint.Ports.Length == 0)
```
hsubramanianaks, line 429: @ababilone Thanks for this PR, can we add a check ```if (endpoint.ports == null || endpoint.Ports.Length == 0)``` just in case if the ports element is null?

hsubramanianaks (approved):
looks good to me. @ababilone Thanks for your first contribution. Appreciate it.

Tatsinnit:
Nice @hsubramanianaks fyi :) I have updated this branch, if this looks cool to you, we could take action on this PR as well. Thanks @ababilone for the contribution. I just noticed to ping you both. Thanks both.

PR #354, opened 2024-01-22
removing dotnet pack for win-arm64

hsubramanianaks:
This PR is removing the step which does dotnet pack for win-arm64 due to vs extension is not ready to consume these arm64 binaries.


Tatsinnit (approved):
☕️🎉

PR #643, opened 2024-08-18
when cloning the pod spec, reserve the previous env variables.

GKilm:
To solve the problem that the element with environment variable in the pod can be available.


GKilm:

